PR #21, opened 2021-03-11
Navigation UI

TheImpressiveDonut:

PR #45, opened 2021-03-15
Event model

hugolan:


app/build.gradle
```diff
@@ -39,9 +39,15 @@ dependencies {
...
     implementation 'com.google.firebase:firebase-auth'
+    implementation 'androidx.wear:wear:1.0.0'
     testImplementation 'junit:junit:4.+'
     androidTestImplementation 'androidx.test.ext:junit:1.1.2'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.3.0'
+    compileOnly 'com.google.android.wearable:wearable:2.6.0'
+    androidTestImplementation 'androidx.test.espresso:espresso-core:3.3.0'
+    androidTestImplementation 'androidx.test:runner:1.3.0'
+    androidTestImplementation 'androidx.test:rules:1.3.0'
+    androidTestImplementation 'androidx.test.espresso:espresso-intents:3.3.0'
```
Ef55, line 50: Those dependencies shouldn't be added.

app/src/main/AndroidManifest.xml
```diff
@@ -2,13 +2,32 @@
...
+        <!--
+               Set to true if your app is Standalone, that is, it does not require the handheld
+               app to run.
+        -->
+        <meta-data
+            android:name="com.google.android.wearable.standalone"
+            android:value="false" />
+
+
+        <activity android:name=".EventCreator" />
```
Ef55, line 30: Same here: we don't want to use Android Wear.

app/src/main/java/com/github/onedirection/Event.java
```diff
@@ -0,0 +1,103 @@
...
+    final private String name;
+    final private String location;
+    final private String date;
+    final private String start_time;
+    final private String end_time;
+
+
+    public Event(int id,String name, String location, String date, String start_time, String end_time){
+        if(name == null || location == null || date == null || start_time == null || end_time == null){
+            throw new IllegalArgumentException();
```
Ef55, line 15: It would be (imo) cleaner to do
```suggestion
        Objects.requireNonNull(name);
        Objects.requireNonNull(location);
        ...
        Objects.requireNonNull(end_time);

```
(Note: we could even write an utility function to write)
```suggestion
        requireNoneNull(name , location, date, start_time, end_time);
```

app/src/main/java/com/github/onedirection/Event.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    public Event set_name(String new_value){
+        if(new_value ==  null){
+            throw new IllegalArgumentException();
+        }
+        if(new_value == this.name){
+            return this;
+        }
+        return new Event(id,new_value,location,date,start_time,end_time);
```
Ef55, line 32: Shorter
```suggestion
        Objects.requireNonNull(new_value);
        return new_value == this.name ?
            this:
            new Event(id,new_value,location,date,start_time,end_time);

```
I would at least change the Null check (so that CodeClimate stops complaining).
The same applies to all setters.

app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,9 @@
 <resources>
     <string name="app_name">1Direction</string>
+    <string name="title_activity_main2">MainActivity</string>
+    <!--
+    This string is used for square devices and overridden by hello_world in
+    values-round/strings.xml for round devices.
+    -->
+    <string name="hello_world">Hello Square World!</string>
```
Ef55, line 8: Don't think there is a main2.

app/src/main/res/layout/activity_main.xml
```diff
@@ -6,13 +6,15 @@
...
+        android:onClick="goToEventCreator"
+        android:text="Create Event"
         app:layout_constraintBottom_toBottomOf="parent"
-        app:layout_constraintLeft_toLeftOf="parent"
-        app:layout_constraintRight_toRightOf="parent"
-        app:layout_constraintTop_toTopOf="parent" />
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent" />
 
 </androidx.constraintlayout.widget.ConstraintLayout>
```
Ef55, line 20: The main activity was removed in #21

Ef55 (changes requested):

app/build.gradle
hugolan: I just delete them?

app/src/androidTest/java/com/github/onedirection/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,28 @@
...
+public class ExampleInstrumentedTest {
+    @Test
+    public void useAppContext() {
+        // Context of the app under test.
+        Context appContext = InstrumentationRegistry.getInstrumentation().getTargetContext();
+        assertEquals("com.github.onedirection", appContext.getPackageName());
+    }
+
+
+}
```
Ef55, line 28: You shouldn't add this file back.

app/src/main/AndroidManifest.xml
```diff
@@ -2,17 +2,23 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.github.onedirection">
 
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
+
+    <uses-feature android:name="android.hardware.type.watch" />
+
```
Ef55, line 8: This shouldn't be added (again, android wear).

app/src/main/java/com/github/onedirection/Event.java
```diff
@@ -0,0 +1,81 @@
...
+        this.name = Objects.requireNonNull(name);
+        this.location = Objects.requireNonNull(location);
+        this.date = Objects.requireNonNull(date);
+        this.start_time = Objects.requireNonNull(start_time);
+        this.end_time = Objects.requireNonNull(end_time);
+        this.id = id;
+
+    }
+
+    public Event set_name(String new_value){
```
Ef55, line 26: Java methods' shouldn't use snake_case.
```suggestion
    public Event setSame(String new_value){
```
Same for all setters.

app/src/test/java/com/github/onedirection/EventTest.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Test
+    public void testEventWithNullArgument(){
+        int id = 001;
+        String name = "name";
+        String location = "location";
+        String date = "date";
+        String start_time = "start_time";
+        String end_time = "end_time";
```
Ef55, line 21: Would be better to define all those constants as _final static_ attributs.

Ef55 (changes requested):

Ef55 (approved):

app/src/main/java/com/github/onedirection/Event.java
```diff
@@ -0,0 +1,71 @@
...
+
+class Event {
+
+    final private int id;
+    final private String name;
+    final private String location;
+    final private String date;
+    final private String start_time;
+    final private String end_time;
+
```
justagoodguy-omega, line 13: Should there not be a boolean field for repeat every week and a field for description of the event?

app/src/main/java/com/github/onedirection/Event.java
```diff
@@ -0,0 +1,71 @@
+package com.github.onedirection;
+
+import java.util.Objects;
+
+class Event {
```
Flechman, line 5: Does not implement Storable

app/src/main/java/com/github/onedirection/Event.java
```diff
@@ -0,0 +1,71 @@
+package com.github.onedirection;
+
+import java.util.Objects;
+
+class Event {
+
+    final private int id;
```
Flechman, line 7: id is not an int, should use the Id class

Flechman (changes requested):
The Event model should implement Storable along with the dedicated methods, and to represent the id it should use the class Id dedicated to unique ids.

Flechman (approved):

PR #47, opened 2021-03-18
Login enhancement

TheImpressiveDonut:


app/src/androidTest/java/com/github/onedirection/SignFragmentTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+        onView(withId(R.id.drawer_layout)).perform(DrawerActions.open());
+        onView(withId(R.id.nav_logout)).perform(click());
+        onView(withText("No")).inRoot(isDialog()).check(matches(isDisplayed())).perform(click());
+        onView(withId(R.id.drawer_layout)).perform(DrawerActions.open());
+        onView(withId(R.id.nav_logout)).check(matches(isDisplayed()));
+
+        onView(withId(R.id.drawer_layout)).perform(DrawerActions.open());
+        onView(withId(R.id.nav_logout)).perform(click());
+        onView(withText("Yes")).inRoot(isDialog()).check(matches(isDisplayed())).perform(click());
```
Ef55, line 62: I believe it's "best practice" to have such strings in an external resource file, instead of having them hard-coded.

app/src/main/java/com/github/onedirection/navigation/NavigationActivity.java
```diff
@@ -50,4 +62,31 @@ public boolean onSupportNavigateUp() {
         return NavigationUI.navigateUp(navController, mAppBarConfiguration)
                 || super.onSupportNavigateUp();
     }
+
+    private void logout(MenuItem signMenuItem, MenuItem logoutMenuItem, TextView drawerUsername,
+                        TextView drawerEmail, DrawerLayout drawer) {
+        AlertDialog.Builder confirmationWindows = new AlertDialog.Builder(this)
+                .setTitle("Log out")
+                .setMessage("Are you sure you want to log out ?")
+                .setPositiveButton("Yes", new DialogInterface.OnClickListener() {
```
Ef55, line 71: Same remark with those strings.

app/src/main/java/com/github/onedirection/navigation/fragment/sign/SignFragment.java
```diff
@@ -0,0 +1,177 @@
...
+import com.github.onedirection.authentication.User;
+import com.github.onedirection.navigation.fragment.home.HomeFragment;
+import com.google.android.material.navigation.NavigationView;
+
+import org.w3c.dom.Text;
+
+public class SignFragment extends Fragment {
+
+    private SignViewModel signViewModel;
+    private boolean register;
```
Ef55, line 38: I think they could be final (but that's a guess).

app/src/main/java/com/github/onedirection/navigation/fragment/sign/SignViewModel.java
```diff
@@ -0,0 +1,60 @@
...
+    LiveData<SignFormState> getLoginFormState() {
+        return signFormState;
+    }
+
+    LiveData<User> getUserResult() {
+        return userResult;
+    }
+
+    public void sign(String username, String password, boolean register) {
+        FirebaseAuthentication auth = FirebaseAuthentication.getInstance();
```
Ef55, line 27: Would be better to use
```suggestion
        AuthenticationService auth = AuthenticationService.getDefaultInstance();
```
so there is no direct dependency on the FirebaseAuthentication.

app/src/main/res/values/strings.xml
```diff
@@ -1,27 +1,32 @@
 <resources>
     <string name="app_name">1Direction</string>
     <string name="title_activity_navigation">NavigationActivity</string>
-    <string name="navigation_drawer_open">Open navigation drawer</string>
-    <string name="navigation_drawer_close">Close navigation drawer</string>
 
-    <string name="nav_header_title">Example Name</string>
-    <string name="nav_header_subtitle">example.name@somewhere.com</string>
+    <string name="nav_header_username">Guest</string>
+    <string name="nav_header_email">empty@email.com</string>
```
Ef55, line 6: I personally think that it would be better to display no email than a placeholder email when the user is a guest.

Ef55 (approved):

app/src/main/res/values/strings.xml
TheImpressiveDonut: I agree, I've changed it

app/src/main/java/com/github/onedirection/navigation/fragment/sign/SignViewModel.java
TheImpressiveDonut: True, I've change it

app/src/main/java/com/github/onedirection/navigation/fragment/sign/SignFragment.java
TheImpressiveDonut: Well not really since register changes multiples times later on and for the signViewModel I would have to add a constructor but I am not really sure how this would work with android

app/src/androidTest/java/com/github/onedirection/SignFragmentTest.java
TheImpressiveDonut: Done

app/src/main/java/com/github/onedirection/navigation/NavigationActivity.java
TheImpressiveDonut: Done

justagoodguy-omega (approved):

app/src/main/java/com/github/onedirection/navigation/fragment/sign/SignFragment.java
Ef55: Indeed, you're correct. I didn't realize that signViewModel was assigned inside the onCreate method, and not in the constructor.

PR #66, opened 2021-03-21
Allow Event to be location-free

Ef55:


Flechman (approved):
Clean code, deactivated code climate but ok

Sosol02 (approved):
We will have to address the cirrus warnings but otherwise LGTM.

PR #69, opened 2021-03-24
Map display

TheImpressiveDonut:


app/src/main/res/values/mapbox_api_keys.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="mapbox_access_token">pk.eyJ1IjoiYXVuZWRpcmVjdGlvbiIsImEiOiJja21ub2dkMDQxd3E4MnZxb3hsbmVnM3FyIn0.gQGoin_4KAfoSeKruQ17hg</string>
```
Sosol02, line 3: Is that a secure token? If it is, maybe we should not store it on git.

app/src/main/java/com/github/onedirection/navigation/fragment/map/MapFragment.java
```diff
@@ -1,45 +1,54 @@
...
-            public void onChanged(String s) {
-                mText.setText(s);
+            public void onMapReady(@NonNull MapboxMap mapboxMap) {
+
+                mapboxMap.setStyle(Style.MAPBOX_STREETS, new Style.OnStyleLoaded() {
+                    @Override
+                    public void onStyleLoaded(@NonNull Style style) {
+
+                    }
+                });
```
Ef55, line 42: This could be a lambda :)

app/src/main/res/values/mapbox_api_keys.xml
TheImpressiveDonut: It is a public key so, it's okay

Sosol02 (approved):
LGTM

Ef55 (approved):

Ef55:
Also (my original comment got lost): the linter thinks that we should update some libs.

app/src/main/java/com/github/onedirection/navigation/fragment/map/MapFragment.java
TheImpressiveDonut: Done

PR #85, opened 2021-03-28
Reverse geocoding

Ef55:


justagoodguy-omega (approved):
Great code what can I say

app/src/main/java/com/github/onedirection/geocoding/NominatimGeocoding.java
```diff
@@ -68,8 +83,32 @@ private static String generateRequestURL(String query){
...
+        );
+        requestQueue.add(request);
+        return result;
+    }
+
     @Override
     public CompletableFuture<NamedCoordinates> getBestNamedCoordinates(String locationName) {
-        return Monads.flatten(sendRequest(locationName).thenApply(NominatimGeocoding::parseResult));
+        return Monads.flatten(
+                sendArrayRequest(generateSearchRequestURL(locationName)).thenApply(r -> parseResult(r, 0))
```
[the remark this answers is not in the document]
TheImpressiveDonut: odersky

app/src/main/java/com/github/onedirection/geocoding/NominatimGeocoding.java
```diff
@@ -36,25 +39,37 @@ public NominatimGeocoding(Context ctx){
...
+        }
+        catch(Exception e){
+            Log.d(LOGCAT_TAG, "Parsing failed");
+            return Optional.empty();
+        }
     }
 
-    private static Optional<NamedCoordinates> parseResult(JSONArray json){
-        Log.d(LOGCAT_TAG, "Parsing: " + json);
+    private static Optional<NamedCoordinates> parseResult(JSONArray json, int idx){
```
TheImpressiveDonut, line 61: You only call this private method with idx = 0, so idx is not really needed. But I suppose this method maybe used for other things in the future?

TheImpressiveDonut (commented):
One question on a private method

app/src/main/java/com/github/onedirection/geocoding/NominatimGeocoding.java
Ef55: This arg specifies which result to retrieve. Currently, I only use the first one (`idx = 0`), so it is indeed useless.
However, adding this arg didn't take much time (as you may have guessed), and leaving the choice to the user to select which result to use is something we might add later, so I already allowed that in this particular method.

TheImpressiveDonut (approved):

PR #90, opened 2021-04-03
Eventqueries

Flechman:
Added queries for events to the database, such as querying events based on their timeframe.
Also added some methods in ConcreteDatabase that facilitate the implementation of event queries.


app/src/main/java/com/github/onedirection/EventQueries.java
```diff
@@ -13,20 +26,73 @@ public EventQueries(Database db) {
         this.db = db;
     }
 
-    public List<Event> getEventsByDay(ZonedDateTime day) {
-        //TODO
-        return null;
+    public CompletableFuture<List<Event>> getEventsInTimeframe(ZonedDateTime start, ZonedDateTime end) {
+        Objects.requireNonNull(start);
+        Objects.requireNonNull(end);
+        if(db.getClass() == ConcreteDatabase.class) {
```
justagoodguy-omega, line 32: If you instead checked that db.getClass != ConcreteDatabase.class and right away throw the exception you could take this whole block out of the if statement. Yeah I do desperately need to write review comments for my sdp grade how did you notice?

app/src/main/java/com/github/onedirection/EventQueries.java
```diff
@@ -13,20 +26,73 @@ public EventQueries(Database db) {
...
+                .with(TemporalAdjusters.next(WeekFields.of(Locale.getDefault()).getFirstDayOfWeek())).minusWeeks(1);
+    }
 
+    public static ZonedDateTime truncateTimeToDays(ZonedDateTime time) {
+        return Objects.requireNonNull(time).truncatedTo(ChronoUnit.DAYS);
+    }
+
+    public static ZonedDateTime truncateTimeToMonths(ZonedDateTime time) {
+        return Objects.requireNonNull(time).truncatedTo(ChronoUnit.DAYS).withDayOfMonth(1);
+    }
```
justagoodguy-omega, line 97: Maybe these should be in a separate time utility class since it isn't really specific to Event querying

justagoodguy-omega (approved):
That is some quality code

app/src/main/java/com/github/onedirection/EventQueries.java
Flechman: They are specific to timeframes of events. Should I move them to the Event class ?

Sosol02: I think moving the functions to Utils or Event is a good idea ; otherwise make these functions private. Having them in the public interface is awkward imo.

Sosol02 (approved):
LGTM

PR #104, opened 2021-04-06
Make geocoding return multiple matches instead of just the best

Ef55:


app/src/main/java/com/github/onedirection/geocoding/GeocodingService.java
```diff
@@ -16,7 +16,15 @@
      * @return The coordinates if the location is found.
      */
     default CompletableFuture<Coordinates> getBestCoordinates(String locationName){
```
Flechman, line 18: Shouldn't you need to check for null ?

app/src/main/java/com/github/onedirection/geocoding/GeocodingService.java
```diff
@@ -16,7 +16,15 @@
...
-        return getBestNamedCoordinates(locationName).thenApply(p -> p.dropName());
+        return getBestNamedCoordinates(locationName).thenApply(NamedCoordinates::dropName);
+    }
+
+    /**
+     * @param locationName Some location's name.
+     * @return A list of its possible coordinates.
+     */
+    default CompletableFuture<List<Coordinates>> getCoordinates(String locationName, int count){
```
Flechman, line 26: Check for null ?

app/src/main/java/com/github/onedirection/utils/Monads.java
```diff
@@ -16,13 +19,17 @@ private Monads() {}
 
     public static <T> CompletableFuture<T> toFuture(Task<T> m){
         CompletableFuture<T> result = new CompletableFuture<>();
-        m.addOnSuccessListener(r -> result.complete(r))
-            .addOnFailureListener(t -> result.completeExceptionally(t))
+        m.addOnSuccessListener(result::complete)
```
Flechman, line 22: Scala-like ❤️

Flechman (approved):
Very clean and concise code

app/src/main/java/com/github/onedirection/geocoding/GeocodingService.java
Ef55: There is no need here; the argument is just passed to another function.

app/src/main/java/com/github/onedirection/geocoding/GeocodingService.java
Ef55: There is no need here; the argument is just passed to another function.

TheImpressiveDonut (approved):
Looks good to me

PR #105, opened 2021-04-12
Link database with EventCreator (and Calendar)

Ef55:


Ef55:
Note: the drop of coverage comes from the fact that the whole notifications was being tested through the EventCreator, which now doesn't use it anymore, so the notification currently has no tests at all.

Ef55:
The events are currently not displayed in the calendar.
@justagoodguy-omega is working on a calendar overhaul on the side, so I won't look into it in this PR.

hugolan (approved):

app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CalendarFragment.java
```diff
@@ -41,16 +44,23 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
...
+           }
+        });
     }
 
-    public void addEventToCalendar(com.github.onedirection.events.Event event, long timeInMillis) {
-        Objects.requireNonNull(event, "tried to add null Event");
-        Event calendarEvent = new Event(timeInMillis, event.getName());
+    public void addEventToCalendar(com.github.onedirection.events.Event event) {
+        Objects.requireNonNull(event, "Tried to add null Event");
+        Event calendarEvent = new Event(event.getStartTime().toInstant().toEpochMilli(), event.getName());
```
Flechman, line 58: Shouldn't it be `event.getStartTime().toEpochSecond()` ? First argument in new Event(...)

Flechman (approved):

app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CalendarFragment.java
Ef55: You are possibly right (note that this Event is **not** one of our Event, but an Event from the old calendar library).
As the mentioned above, the whole calendar is getting revamped in PR #111, so I just left my code related to the calendar as-is, and didn't bother too much with the code you highlighted (which I should most likely not even have included in this PR).

PR #108, opened 2021-04-14
My location button

TheImpressiveDonut:


app/src/androidTest/java/com/github/onedirection/map/MapFragmentTest.java
```diff
@@ -142,6 +168,18 @@ public void testAddEventPutsMarkerOnMap() {
         onView(withId(R.id.fragment_map_event_name)).check(matches(withText(TEST_EVENT_1.getName())));
     }
 
+    @Test
+    @Ignore("Cirrus reject")
+    public void testMyLocationIsAppearing() {
```
Ef55, line 173: Happy to know that I'm not the only one with this kind of problems.

Ef55 (changes requested):
Only one issue: there are some Linter's warnings.
(Which I could not highlight for some reason).

Ef55 (approved):

app/src/main/java/com/github/onedirection/navigation/fragment/map/MapFragment.java
```diff
@@ -38,14 +46,26 @@
...
+
     public static final String SYMBOL_ID = "MARKER_MAP";
+    public static final String MY_LOCATION_ID = "MY_LOCATION_MAP";
 
     @Nullable
     @Override
     public View onCreateView(@NonNull LayoutInflater inflater,
                              @Nullable ViewGroup container,
                              @Nullable Bundle savedInstanceState) {
+        deviceLocationProvider = (DeviceLocationProvider) requireActivity();
```
Sosol02, line 61: I don't like the unconditionnal cast, but i don't know how you could have made it better

Sosol02 (approved):

app/src/main/java/com/github/onedirection/navigation/fragment/map/MapFragment.java
Ef55: We should be able to do better after PR #109

PR #109, opened 2021-04-14
Make DeviceLocationProvider usable in more cases (i.e. out of an activity)

Ef55:


Ef55:
<details>
<summary> The inheritance diagram: </summary>

</details>

app/src/main/java/com/github/onedirection/geolocation/AbstractDeviceLocationProvider.java
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Only abstract method; it is so because this is the only thing we can't handle
+     * generally (we need to be an activity to handle this easily).
+     *
+     * @return Whether the request was granted.
+     * @see DeviceLocationProviderActivity
+     * @see DeviceLocationProviderNoRequests
+     */
+    public abstract CompletableFuture<Boolean> requestFineLocationPermission();
+
```
TheImpressiveDonut, line 105: What about the onRequestPermissionsResult that is in DeviceLocationProviderActivity, should I add the method to the activity where I plan to use AbstractDeviceLocationProvider? Actually, in which way should I use AbstractDeviceLocationProvider?

TheImpressiveDonut (commented):
Only one point to discuss, everything else is fine for me

app/src/main/java/com/github/onedirection/geolocation/AbstractDeviceLocationProvider.java
```diff
@@ -100,45 +68,56 @@ public void onLocationAvailability(@NonNull LocationAvailability locationAvailab
...
-        if(requestCode == R.integer.location_permission_code){
-            for(int i = 0; i < permissions.length; ++i){
-                if(permissions[i].equals(Manifest.permission.ACCESS_FINE_LOCATION) && grantResults[i] == PackageManager.PERMISSION_GRANTED){
-                    permissionRequestResult.complete(true);
-                }
-            }
-        }
+    //************************ Setup/Internal methods *****************************************
+
+    private CompletableFuture<Boolean> createLocationRequest() {
```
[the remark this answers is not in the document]
Ef55: Fair point; I would still keep it, as there is no down-side to having it and the result might be of use if we decide to change some things.

app/src/main/java/com/github/onedirection/geolocation/AbstractDeviceLocationProvider.java
Ef55: You have multiple way to go about this:
- You can implement your own concrete version of it if you have some way of asking the user for permission;
- If you implement an activity, you can inherit DeviceLocationProviderActivity which will do it for you;
- If you have no way of asking (or don't want to), you can simply instantiate LocationProviderNoRequests.
Note it is also possible to do the following: request the permission from user preemptively (e.g. at start of some activity), and later use a LocationProviderNoRequests (since you already know that the permission is granted).

If you want to request permission from a fragment without going the dirty way, there is no clean solution, or at least we didn't find it.

app/src/main/java/com/github/onedirection/geolocation/AbstractDeviceLocationProvider.java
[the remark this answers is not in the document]
TheImpressiveDonut: Okay

app/src/main/java/com/github/onedirection/geolocation/AbstractDeviceLocationProvider.java
TheImpressiveDonut: Okay, thanks

TheImpressiveDonut (approved):

app/src/main/java/com/github/onedirection/geolocation/DeviceLocationProviderActivity.java
```diff
@@ -0,0 +1,109 @@
...
+ * Device location provider which uses the activity extending it to
+ * ask for user's permission to track its location.
+ */
+public abstract class DeviceLocationProviderActivity extends AppCompatActivity implements DeviceLocationProvider {
+    private final static LocationRequest LOCATION_REQUEST = LocationRequest.create();
+
+    static {
+        LOCATION_REQUEST.setInterval(10000);
+        LOCATION_REQUEST.setFastestInterval(5000);
+        LOCATION_REQUEST.setPriority(LocationRequest.PRIORITY_HIGH_ACCURACY);
```
justagoodguy-omega, line 28: Just a detail but we could probably extract this into a method or just put the values into variables

justagoodguy-omega (approved):
Good code

PR #110, opened 2021-04-14
Recurringevents

Flechman:
Extention of the Event model to support recurring events, along with the adaptation of queries from the database to support recurring events.


app/src/main/java/com/github/onedirection/EventQueries.java
```diff
@@ -33,15 +39,42 @@ public EventQueries(Database db) {
             CompletableFuture<List<Event>> res2 = cdb.filterWhereGreaterLessEq(EventStorer.KEY_EPOCH_END_TIME, ti, tf, EventStorer.getInstance()); // ti < eventEndTime <= tf
             CompletableFuture<List<Event>> res3 = cdb.filterWhereLess(EventStorer.KEY_EPOCH_START_TIME, ti, EventStorer.getInstance()); // eventStartTime < ti
             CompletableFuture<List<Event>> res4 = cdb.filterWhereGreater(EventStorer.KEY_EPOCH_END_TIME, tf, EventStorer.getInstance()); // eventEndTime > tf
-            return CompletableFuture.allOf(res1, res2, res3, res4).thenApply(ignoredVoid -> {
+            CompletableFuture<List<Event>> recurringEvents = cdb.filterWhereGreaterEq(EventStorer.KEY_RECURRING_PERIOD, new Long(0), EventStorer.getInstance());
```
Ef55, line 42: [Problem] Listen to the linter: 
```suggestion
            CompletableFuture<List<Event>> recurringEvents = cdb.filterWhereGreaterEq(EventStorer.KEY_RECURRING_PERIOD, Long.valueOf(0), EventStorer.getInstance());
```
In fact, I think that the following would also work
```suggestion
            CompletableFuture<List<Event>> recurringEvents = cdb.filterWhereGreaterEq(EventStorer.KEY_RECURRING_PERIOD, 0L, EventStorer.getInstance());
```
or possibly even
```suggestion
            CompletableFuture<List<Event>> recurringEvents = cdb.filterWhereGreaterEq(EventStorer.KEY_RECURRING_PERIOD, 0, EventStorer.getInstance());
```
(I don't exactly remember how Java behaves with implicit conversions + boxing).

app/src/main/java/com/github/onedirection/EventQueries.java
```diff
@@ -25,6 +28,9 @@ public EventQueries(Database db) {
     public CompletableFuture<List<Event>> getEventsInTimeframe(ZonedDateTime start, ZonedDateTime end) {
         Objects.requireNonNull(start);
         Objects.requireNonNull(end);
+        if(end.toEpochSecond() - start.toEpochSecond() <= 0) {
+            return CompletableFuture.completedFuture(new ArrayList<Event>());
+        }
         if(db.getClass() == ConcreteDatabase.class) {
```
Ef55, line 34: [Detail] That's dirty :) (I mean the type test)

app/src/main/java/com/github/onedirection/EventQueries.java
```diff
@@ -33,15 +39,42 @@ public EventQueries(Database db) {
...
+                for(Event e : recurring) {
+                    long tStart = e.getStartTime().toEpochSecond();
+                    long tEnd = e.getEndTime().toEpochSecond();
+                    long duration = tEnd-tStart;
+                    long period = e.getRecurringPeriod().get().getEpochSecond();
+                    long x = Math.max(1, ti/tStart);
+                    while((tStart+(x-1)*period) < tf) {
+                        long startTime = (tStart+(x-1)*period);
+                        long endTime = startTime + duration;
+                        Event newEvent = new Event(Id.generateRandom(), e.getName(), e.getLocationName(), e.getCoordinates(),
```
Ef55, line 64: [Problem] The fact that you generate a random Id for each "instantiation" of the Event seems like it could cause lots of bug.
You should add to the documentation of Event what is the expected update mechanism for recurring event.

What happens if I modify one of the instances ? Does it change all instances ? Do I get one simple event and a recurring one ?

app/src/main/java/com/github/onedirection/EventQueries.java
```diff
@@ -78,17 +111,4 @@ public EventQueries(Database db) {
...
-                .with(TemporalAdjusters.next(WeekFields.of(Locale.getDefault()).getFirstDayOfWeek())).minusWeeks(1);
-    }
-
-    public static ZonedDateTime truncateTimeToDays(ZonedDateTime time) {
-        return Objects.requireNonNull(time).truncatedTo(ChronoUnit.DAYS);
-    }
-
-    public static ZonedDateTime truncateTimeToMonths(ZonedDateTime time) {
-        return Objects.requireNonNull(time).truncatedTo(ChronoUnit.DAYS).withDayOfMonth(1);
-    }
```
Ef55, line 93: Nice clean up of the interface 👌

app/src/main/java/com/github/onedirection/database/store/EventStorer.java
```diff
@@ -65,17 +66,31 @@ public Event mapToStorable(Map<String, Object> m) {
...
-                    ZonedDateTime.ofInstant(Instant.ofEpochSecond(epochEndTime), ZoneId.systemDefault()));
+            if(recurringPeriod == null) {
+                return new Event(new Id(UUID.fromString(id)), name, locationName, new Coordinates(coordLatitude, coordLongitude),
+                        TimeUtils.epochToZonedDateTime(epochStartTime),
+                        TimeUtils.epochToZonedDateTime(epochEndTime));
+            } else {
+                return new Event(new Id(UUID.fromString(id)), name, locationName, new Coordinates(coordLatitude, coordLongitude),
+                        TimeUtils.epochToZonedDateTime(epochStartTime),
+                        TimeUtils.epochToZonedDateTime(epochEndTime), Instant.ofEpochSecond(recurringPeriod));
+            }
```
Ef55, line 93: [Detail] There should be a better/shorter way to do this using _Optional_ (might need to change Event's constructors a bit tho).

app/src/main/java/com/github/onedirection/events/Event.java
```diff
@@ -25,31 +30,40 @@
     final private Coordinates location;
     final private ZonedDateTime startTime;
     final private ZonedDateTime endTime;
+    final private Instant recurringPeriod;
```
Ef55, line 33: [Problem] Instant is not a suitable class to represent the time step between occurrences ([This class models a **single instantaneous** point on the time-line](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/Instant.html)).
I think that [Duration](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/Duration.html) is a better fit (not sure if it is the best tho).

app/src/main/java/com/github/onedirection/events/Event.java
```diff
@@ -131,6 +152,14 @@ public Duration getDuration() {
         return Duration.between(startTime, endTime);
     }
 
+    public boolean isRecurrent() {
+        return getRecurringPeriod().isPresent();
+    }
+
+    public Optional<Instant> getRecurringPeriod() {
```
Ef55, line 159: [Nitpick] It seems to me that "Recurring period" is not correct [English](https://www.deepl.com/translator#fr/en/p%C3%A9riode%20de%20r%C3%A9currence).

Ef55 (changes requested):

app/src/main/java/com/github/onedirection/EventQueries.java
Flechman: Generating a random Id for each "instantiation" is intentional : When the recurring event is loaded from the database, it may generate more events (recurrence of the event that has been loaded from the db) that do not exist is the db, therefore they "need" to be assigned an individual Id. If one of the newly generated events is modified, then the change should only apply to this one and not the others  of the recurring series, so we are going to store it as a new event that is not recurring.
Another thing that I should change is that these newly generated events should not have a "recurring" attribute (this follows from the reasoning above with the modification of one).

Flechman: After further thinking about the potential issues of the way it is done, I will keep the "recurring" attribute, and give the same Ids to every event that is part of the same recurring series of events. The reason for the latter is that in order to modify the recurring event series, we need to change the "root" event which is the one that is stored in the database. So we need to know the Id of this "root" event. Keeping the "recurring" attribute is necessary to identify the Id as being recurring.
A limitation of the implementation is that we cannot modify just a single event from the recurring event series while still keeping it in the recurring series. Modifying an event that belongs to a series of recurring events means modifying the  whole series not just this one. If we really want the feature of changing just this event even though it's part of a series of events then we'll  have to create a new issue.

Ef55: I'm with you on this one. We can always change the implementation later if needed.
With the one you ended with, at least things stay simple and seem to cause less problems.

hugolan (approved):

app/src/main/java/com/github/onedirection/events/Event.java
Flechman: I will create a new PR for this one

app/src/main/java/com/github/onedirection/database/store/EventStorer.java
```diff
@@ -64,19 +66,17 @@ public Event mapToStorable(Map<String, Object> m) {
         String name = (String) m.get(KEY_NAME);
         Double coordLatitude = (Double) m.getOrDefault(KEY_COORD_LATITUDE, null);
         Double coordLongitude = (Double) m.getOrDefault(KEY_COORD_LONGITUDE, null);
+        Coordinates coords = coordLatitude == null || coordLongitude == null ? null : new Coordinates(coordLatitude, coordLongitude);
         String locationName = (String) m.get(KEY_COORD_NAME);
+
         long epochStartTime = (long) Objects.requireNonNull(m.get(KEY_EPOCH_START_TIME));
         long epochEndTime = (long) Objects.requireNonNull(m.get(KEY_EPOCH_END_TIME));
+        Long recurringPeriod = (Long) m.getOrDefault(KEY_RECURRING_PERIOD, null);
```
Ef55, line 74: [Nitpick] This should be long :)

Ef55 (approved):
If you fix the _Instant_ issue in a future PR (as it isn't a bug per se), everything is good.

PR #113, opened 2021-04-16
Sol notifications when event will happen

Sosol02:
This is a draft PR to allow  you to see and comment but im not finished _at all_.


Sosol02:
This PR f i n a l l y works but it still has a few TODOs to reslve (mostly via testing) and i want to refactor the schedule methode to take a ZonedDateTime. Otherwise, I also  need to add tests :) géantvidemourir

Sosol02:
I did my best, plz review uwu

app/src/main/java/com/github/onedirection/notification/NotificationPublisher.java
```diff
@@ -0,0 +1,99 @@
...
+                        break;
+                    }
+                    // e is within lowerBound..upperBound
+                    Notification notif = notifications.getNotification(e, appContext);
+                    int notifId = NotificationIdGenerator.getUniqueId();
+                    Log.d(LOG_TAG, "Send notifs: notif: " + notif + ", id: " + notifId);
+                    nm.notify(notifId, notif);
+                }
+                if (!scheduledNextTime) {
+                    Log.d(LOG_TAG, "No event was available for scheduling the next call, " +
```
Ef55, line 92: [Nitpick] It would be better to use [String::format](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#format(java.lang.String,java.lang.Object...)).

app/src/main/java/com/github/onedirection/notification/Notifications.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+        return global;
+    }
+
+    private Notifications(Context context) {
+        setupNotifChannel(context);
+    }
+
+    private void setupNotifChannel(Context context) {
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.O) {
```
Ef55, line 46: What happens if the version is not the expected one ?

app/src/main/java/com/github/onedirection/notification/Notifications.java
```diff
@@ -0,0 +1,86 @@
...
+    void schedule(Context context, ZonedDateTime when, BroadcastReceiver handler) {
+        ZonedDateTime now = ZonedDateTime.now();
+        if (when.isBefore(now)) {
+            Log.d("schedule", "Registering handler with due date in the past: now: " + now + ", when: " + when);
+        }
+
+        Intent intent = new Intent(context, handler.getClass());
+        // check what requestCode does. used to be notifId.
+        // https://stackoverflow.com/questions/21526319/whats-requestcode-used-for-on-pendingintent
+        // actually seems to not really matter for our usecase.
```
Ef55, line 80: You should probably fix this (or at least use a TODO).

Ef55 (changes requested):

app/src/main/java/com/github/onedirection/EventsView.java
```diff
@@ -36,6 +35,6 @@ protected void onCreate(Bundle savedInstanceState) {
 
         HashSet<Event> eventHashSet = new HashSet<>();
         eventHashSet.add(event);
-        Notifications.getInstance(getApplicationContext()).scheduleEventNotifs(getApplicationContext(), eventHashSet);
+        //Notifications.getInstance(getApplicationContext()).scheduleEventNotifs(getApplicationContext(), eventHashSet);
```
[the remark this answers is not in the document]
Sosol02: you are right uwu

app/src/main/java/com/github/onedirection/navigation/NavigationActivity.java
```diff
@@ -65,6 +66,10 @@ public boolean onMenuItemClick(MenuItem menuItem) {
                 return false;
             }
         });
+
+        // Starts the notifications handler.
+        // DO NOT REMOVE THIS LINE otherwise notifications wont work!
+        Notifications.getInstance(getApplicationContext());
```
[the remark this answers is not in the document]
Sosol02: will do that

app/src/main/java/com/github/onedirection/notification/NotificationPublisher.java
```diff
@@ -0,0 +1,99 @@
...
+ * that will recall this function, etc.
+ */
+public class NotificationPublisher extends BroadcastReceiver {
+
+    private static final String LOG_TAG = "NotificationPublisher";
+    /**
+     * Delta time in seconds within which we consider an event to trigger.
+     * Consider looking at onRecieve to get the full picture.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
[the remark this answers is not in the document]
Sosol02: i keep that to make my purposes explict

app/src/main/java/com/github/onedirection/notification/NotificationPublisher.java
Sosol02: Yes but it is only for 1 element so i think it's ok

app/src/main/java/com/github/onedirection/notification/Notifications.java
Sosol02: This is only necessary on modern android versions. Code taken straight from the documentation.

app/src/main/java/com/github/onedirection/notification/Notifications.java
Sosol02: This used to be a TODO but i checked the docs and the requestCode doesnt matter.

app/src/main/java/com/github/onedirection/navigation/NavigationActivity.java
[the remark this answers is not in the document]
Sosol02: Actually the Android docs recommand not extending Application. The SO you linked dates from 2012.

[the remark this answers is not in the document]
Ef55: Fair point, didn't realized when I found the post.
It seems we could use the [startup library](https://developer.android.com/reference/androidx/startup/package-summary) instead, but this goes out of the scope of this PR imo.

Ef55 (approved):
Lgtm

(thank's for setting *startup* up).

app/src/main/java/com/github/onedirection/notification/Notifications.java
```diff
@@ -0,0 +1,88 @@
...
+            NotificationChannel channel = new NotificationChannel(CHANNEL_ID, CHANNEL_NAME, importance);
+            channel.setDescription(CHANNEL_DESC);
+            NotificationManager notificationManager = context.getSystemService(NotificationManager.class);
+            notificationManager.createNotificationChannel(channel);
+        }
+        schedule(context, ZonedDateTime.now().plusSeconds(3), new NotificationPublisher());
+    }
+    
+    Notification getNotification(Event event, Context context) {
+        String text =
```
hugolan, line 60: maybe use stringbuilder

hugolan (approved):

PR #126, opened 2021-04-16
Got rid of the code smell in EventQueries, and added some method headings in the Database interface

Flechman:
The code smell in EventQueries is the `if(db.getClass() == ConcreteDatabase.class)`. The methods added in the Database interface are there to remedy the code smell.


Ef55 (approved):
I only have one remark, which isn't really linked to this PR:

Using [_Monads::toFuture_](https://github.com/Sosol02/sdp/blob/main/app/src/main/java/com/github/onedirection/utils/Monads.java) might make _completeOnList_ shorter and easier to read.
Also, it could be static.

Sosol02 (approved):
lgtm

PR #127, opened 2021-04-18
Recurring events revisited

Flechman:
Revisited the way recurring events work and are stored.
There is now a specific model called "Recurring" which stores information about the recurrence of an event. Each event part of a recurrence series has a Recurring model and and has pointers to the previous event and the next event in the recurrence series.
This enables more freedom to the modification of any event whether it is recurring or not, + it keeps a better generalization of the event model in that generic operations on events do not have to distinguish between a recurring and a non-recurring event.


app/src/main/java/com/github/onedirection/events/Event.java
```diff
@@ -85,11 +79,11 @@ public Event(Id id, String name, NamedCoordinates location, ZonedDateTime startT
         this(id, name, Objects.requireNonNull(location).name, Optional.of(location.dropName()), startTime, endTime, Optional.empty());
     }
 
-    public Event(Id id, String name, NamedCoordinates location, ZonedDateTime startTime, ZonedDateTime endTime, Instant recurringPeriod) throws IllegalArgumentException {
+    public Event(Id id, String name, NamedCoordinates location, ZonedDateTime startTime, ZonedDateTime endTime, Recurrence recurringPeriod) throws IllegalArgumentException {
         this(id, name, Objects.requireNonNull(location).name, Optional.of(location.dropName()), startTime, endTime, Optional.of(recurringPeriod));
     }
```
justagoodguy-omega, line 84: Is this constructor useful since the one underneath does the same but with Optional<Recurrence>? I mean if there is a version with Optional why is there a version without?

justagoodguy-omega (commented):
So I'm just wondering now that you have this linked list implementation, if we give a specific day and month and year, how will you operate to find if one of these recurring events has an event on that day? Will you have to go through the whole list of every recurring event?

app/src/main/java/com/github/onedirection/events/Recurrence.java
```diff
@@ -0,0 +1,102 @@
...
+@Immutable
+/**
+ * Class representing the recurrence of an event.
+ * A recurring event is stored as a recurrence series where all its recurring times are stored as individual events.
+ * The recurrence series is abstracted as a linked-list of events, where each event in the linked-list has pointers to its neighboring events. The Recurrence class is thus here to append these informations
+ * to an event whenever this event belongs to a recurrence series.
+ * INVARIANT : A recurrence series has always more than 1 element (event). That is, if an event belongs to a recurrence series (i.e. is recurrent) then its pointers to the previous and next events
+ * of the recurrence series are not both null.
+ */
+public class Recurrence {
```
Ef55, line 20: This should be serializable. 
In fact, I'm a bit worried by the fact that tests pass without this class being serializable...

app/src/main/java/com/github/onedirection/events/Recurrence.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private Id groupId;
+    private Duration period;
+    private Id prevEvent;
+    private Id nextEvent;
```
Ef55, line 25: What do those attributs represent ?
Is groupId an Id for the whole "set" of events, or for the recurrence of this particular instance of the event ?
Are the other Ids the one of the instances, or of the recurrences linked to those Ids ?
What is this period ? The offset from the the time of the "initial" event, or the offset from the previous instance ?

app/src/main/java/com/github/onedirection/events/Recurrence.java
```diff
@@ -0,0 +1,102 @@
...
+        return Objects.requireNonNull(newId).equals(groupId) ? this : new Recurrence(newId, period, Optional.ofNullable(prevEvent), Optional.ofNullable(nextEvent));
+    }
+
+    public Recurrence setPrevEvent(Optional<Id> newId) {
+        return Objects.requireNonNull(newId).equals(Optional.ofNullable(prevEvent)) ? this : new Recurrence(groupId, period, newId, Optional.ofNullable(nextEvent));
+    }
+
+    public Recurrence setNextEvent(Optional<Id> newId) {
+        return Objects.requireNonNull(newId).equals(Optional.ofNullable(nextEvent)) ? this : new Recurrence(groupId, period, Optional.ofNullable(prevEvent), newId);
+    }
```
Ef55, line 69: I'm not sure those are useful.
Also, what happens if there is already a next and I replace it ?
What once I store the event in the DB ?

app/src/main/java/com/github/onedirection/events/Recurrence.java
```diff
@@ -0,0 +1,102 @@
...
+        return Optional.ofNullable(nextEvent);
+    }
+
+    public Duration getPeriod() {
+        return period;
+    }
+
+    public Recurrence setGroupId(Id newId) {
+        return Objects.requireNonNull(newId).equals(groupId) ? this : new Recurrence(newId, period, Optional.ofNullable(prevEvent), Optional.ofNullable(nextEvent));
+    }
```
Ef55, line 61: Why can we replace the ID ?
What's the point ? Won't it cause any issue with the DB ?

Ef55 (changes requested):
The code is good but I'm a bit puzzled about the hows and whys.

There is at least missing documentation, and depending on it, some things might not be as practical as they should.

A good start to help me understand would be to answer:
"I want to create an event recurring every week for three weeks. How do I do that ?"

Flechman:
> So I'm just wondering now that you have this linked list implementation, if we give a specific day and month and year, how will you operate to find if one of these recurring events has an event on that day? Will you have to go through the whole list of every recurring event?

Since every event from the linked list is stored into the database, no additional work is needed. Just call the methods implemented in EventQueries

app/src/main/java/com/github/onedirection/events/Event.java
Flechman: Having one with Optional enables to pass null elements wrapped in the Optional (it is possible, ex. with a null location). Having another without  Optional enables to abstract away the Optional wrapper.

app/src/main/java/com/github/onedirection/events/Recurrence.java
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public Recurrence(Id groupId, Duration period, Optional<Id> prev, Optional<Id> next) {
+        this.groupId = Objects.requireNonNull(groupId);
+        this.period = Objects.requireNonNull(period);
+        this.prevEvent = Objects.requireNonNull(prev).orElse(null);
+        this.nextEvent = Objects.requireNonNull(next).orElse(null);
+    }
+
+    public Recurrence(Id groupId, long periodSeconds, Optional<Id> prev, Optional<Id> next) {
+        this(groupId, Duration.ofSeconds(periodSeconds), prev, next);
```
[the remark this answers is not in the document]
Flechman: Where do I say that ? It's false, they can be both null.

app/src/main/java/com/github/onedirection/events/Recurrence.java
Flechman: For the testing part, I implemented the testing of Recurrence in the exact same way as the testing for Event was made.

Ef55: Yeah I realized afterwards that we never really test serialization, and since EventCreator doesn't use recurring events yet, the problem doesn't manifest itself.

app/src/main/java/com/github/onedirection/events/Recurrence.java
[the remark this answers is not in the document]
justagoodguy-omega: image

app/src/main/java/com/github/onedirection/events/Event.java
justagoodguy-omega: image

app/src/main/java/com/github/onedirection/events/Recurrence.java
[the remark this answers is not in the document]
Ef55: I guess that both might be null, but both cannot be null at the same time (if you see what I mean).

justagoodguy-omega:
> Since every event from the linked list is stored into the database, no additional work is needed. Just call the methods implemented in EventQueries

Sorry if I'm not understanding something but given that a recurring event can happen a potentially infinite number of times how can they all be in the database?

app/src/main/java/com/github/onedirection/events/Recurrence.java
Flechman: I'll nevertheless make it serializable

app/src/main/java/com/github/onedirection/events/Recurrence.java
Flechman: You're right, this operation doesn't really fit with the purpose of the class Recurrence

app/src/main/java/com/github/onedirection/events/Recurrence.java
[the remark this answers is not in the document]
Flechman: Sorry my explanations are not very clear. 
The Recurrence class is an object that is appended to an Event whenever this event belongs to a recurrence series, but it is not the recurrence series itself. 
The Recurrence object appended to an Event thus provides the information of this event in the recurrence series.
In the Recurrence object, we authorize the fact that both prev AND next can be null at the same time (without talking about recurrence series yet). This is useful for example when we want to create a recurrence series : upon creation of the recurrence series we base ourselves on an Event that is our reference for each Event  in our series. This Event has a Recurrence (R) appended to it to show that it IS a recurring event. But since the recurrence series has not been created yet, the pointers to other Event ids in the Recurrence (R) are null for the moment.
Now, the invariant applies to a recurrence series, i.e. to our linked list of Events that have Recurrence objects appended to them. The invariant says that the size of this linked list should not be less or equal to 1, i.e. there  should always be  at least 2 events in the linked-list. Upon creation of a recurrence series, if we see that the series will contain less than 2 events, then we don't create a recurrence series and we just create an ordinary Event (that doesn't have a Recurrence object  appended to it since it does not belong to a recurrence series).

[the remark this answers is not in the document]
Flechman: The recurrence series is defined and "recognized" with the attribute _groupId_ in Recurrence.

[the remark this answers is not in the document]
justagoodguy-omega: I see thanks! :)

app/src/main/java/com/github/onedirection/EventQueries.java
```diff
@@ -43,73 +55,376 @@ public EventQueries(Database db) {
...
+
+    /**
+     * Method to create a full series of recurring events based on the attributes of a given event. The events are recurring from the startTime of the given event to time 'endTime' (included)
+     * given in the Recurrence attribute of 'event', and the time between each recurrence of the event is based on the period given in the Recurrence attribute of 'event'.
+     * Ex use : addRecurringEvent(event("tennis", startTime: "Jan 1 2021).Recurrence(period:"weekly", endTime: " Jan 1 2022")) means the event "tennis" will occur every week
+     * starting at the begining of 2021 and will end at the begining of 2022
+     * @param event (Event) : the given event representing the recurrence series
+     * @return (CompletableFuture<Integer>) : If the recurrence series has been successfully created and stored into the database it returns the number of stored events, else it returns 0
+     */
+    public CompletableFuture<Integer> addRecurringEvent(Event event) {
```
Ef55, line 137: [Opinion] Imo, it would be nice to have a simpler wrapper around those two:
`
public CompletableFuture<Integer> addEvent(Event event) {
  if(event.isRecurrent()){
    addRecuringEvent(event);
  }
  else{
    addNonRecurringEvent(event);
  }
}
`
But I can add it myself in a future PR (if you don't see any issue to that).

app/src/main/java/com/github/onedirection/EventQueries.java
```diff
@@ -43,73 +55,376 @@ public EventQueries(Database db) {
...
+    public CompletableFuture<Id> modifyEvent(Event event) {
+        return db.retrieve(Objects.requireNonNull(event).getId(), event.storer()).thenCompose(e -> {
+            Objects.requireNonNull(e, "Event doesn't exist in the database");
+
+            if(!e.getRecurrence().equals(event.getRecurrence())) {
+                throw new IllegalArgumentException("Cannot change the recurrence attributes of the event");
+            }
+
+            return e.equals(event) ? CompletableFuture.completedFuture(e.getId())
+                    : db.store(new Event(e.getId(), event.getName(), event.getLocationName(), event.getCoordinates(), event.getStartTime(), event.getEndTime(), e.getRecurrence()));
```
Ef55, line 287: [Opinion] I think that `changeRecurringSeriesEndTime` and `convertToRecurring` should be automatically called when 
`event.getRecurrent().endTime != e.getRecurrence().endTime`
or
`event.getRecurrent() != null && != e.getRecurrence() == null`
(this is not valid code, just conceptual ideas).

Ef55 (commented):
I still have one overall issue:
In my opinion, EventQueries should propose really high level methods, and I fell like there should be some methods whose only point is to check properties of an event and call appropriate functions, instead of having the user dissociate between all those cases.

I don't know if you feel this should be added in another PR, so I'm waiting for you opinion @Flechman

app/src/main/java/com/github/onedirection/EventQueries.java
Flechman: The problem is that both methods in the wrapper don't return the same type. But  this could be changed later if  this is really needed.

app/src/main/java/com/github/onedirection/EventQueries.java
Flechman: I think it's better for the moment to handle this on the programmer side as these methods are quite costly since they  query the database.

app/src/main/java/com/github/onedirection/events/Recurrence.java
Flechman: I hope it's clearer now. We made a call so that I could explain to you the design.

app/src/main/java/com/github/onedirection/events/Recurrence.java
Flechman: After our call I hope this part is clearer as well now.

Ef55 (approved):

justagoodguy-omega (approved):

PR #140, opened 2021-04-23
Update dependencies for compatibility between firestore and mapquest

TheImpressiveDonut:
problem with protobuf package between firestore and mapquest

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -69,9 +69,9 @@ dependencies {
     implementation 'com.mapbox.mapboxsdk:mapbox-android-sdk:9.6.1'
     implementation 'com.mapbox.mapboxsdk:mapbox-android-plugin-annotation-v9:0.8.0'
     implementation 'com.mapquest:navigation:3.5.0'
-    implementation('com.mapquest:mapping-android-sdk:2.0.10') {
+    implementation('com.mapquest:mapping-android-sdk:3.0.2') { //this library might not be needed
         exclude group: 'com.android.support'
-        exclude group: 'com.android.util'
+        exclude group: 'com.mapbox.mapboxsdk' //change when navigation is done
     }
     implementation 'androidx.coordinatorlayout:coordinatorlayout:1.1.0'
     testImplementation 'junit:junit:4.13.2'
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/onedirection/database/utils/FirebaseUtils.java
```diff
@@ -13,7 +14,13 @@ public static FirebaseFirestore getFirestore() {
         if (firestore == null) {
             firestore = FirebaseFirestore.getInstance();
             if (useEmulator) {
-                firestore.useEmulator("10.0.2.2", 8080);
```
justagoodguy-omega, line 16: Maybe donc hardcode values like this, you can use string.xml instead for example

justagoodguy-omega (changes requested):

justagoodguy-omega (approved):

app/build.gradle
```diff
@@ -47,14 +58,21 @@ dependencies {
     implementation 'androidx.cardview:cardview:1.0.0'
     implementation 'com.google.android.material:material:1.4.0-alpha02'
     implementation 'androidx.annotation:annotation:1.2.0'
-    implementation platform('com.google.firebase:firebase-bom:26.6.0')
-    implementation 'com.google.firebase:firebase-auth'
-    implementation 'com.google.firebase:firebase-firestore'
+    // firebases library version are changed to be compatible with mapquest
+    //noinspection GradleDependency
+    implementation 'com.google.firebase:firebase-auth:20.0.3'
+    //noinspection GradleDependency
```
Ef55, line 64: Leftovers. Or does gradle read comments ?

Ef55 (changes requested):

app/build.gradle
TheImpressiveDonut: So the noinspection is for the lint warning since there are new version available

TheImpressiveDonut: For gradle it is just a comment, it's an instruction for the linter

Ef55: So it does indeed read the comments (but not gradle).
Thanks

Ef55 (approved):

PR #141, opened 2021-04-26
Reorganize geolocation's packaging structure

Ef55:
The geolocation package grew bigger than expected during last week, and will/would grow even more this week.
So I took the liberty of adding subpackages in it.


Sosol02 (approved):
Much import, such refactor

hugolan (approved):

PR #142, opened 2021-04-26
Implement cached geocoding

Ef55:


app/src/main/java/com/github/onedirection/geolocation/geocoding/GeocodingCache.java
```diff
@@ -0,0 +1,57 @@
...
+import com.github.onedirection.geolocation.Coordinates;
+import com.github.onedirection.geolocation.NamedCoordinates;
+import com.github.onedirection.utils.Cache;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+
+public final class GeocodingCache implements GeocodingService {
+    private static class Key{
```
Sosol02, line 22: There may be a good reason but why not reuse utils.Pair?

Sosol02 (changes requested):

app/src/main/java/com/github/onedirection/geolocation/geocoding/GeocodingCache.java
Ef55: The reason is quite simple: I forgot its existence. 
Gonna fix it right now.

Sosol02 (approved):

app/src/main/java/com/github/onedirection/geolocation/geocoding/NominatimGeocoding.java
```diff
@@ -114,13 +114,6 @@ private static String generateReverseRequestURL(Coordinates coordinates){
         return result;
     }
 
-    @Override
-    public CompletableFuture<NamedCoordinates> getBestNamedCoordinates(String locationName) {
-        return Monads.flatten(
-                sendArrayRequest(generateSearchRequestURL(locationName, 1)).thenApply(r -> parseResult(r, 0))
-        );
-    }
-
```
TheImpressiveDonut, line 123: I'm not really sure how everything works, but I was wondering why did you remove this method here and instead made this method default in the interface GeoService?

TheImpressiveDonut (changes requested):
One point to see, everything else looks good to me

app/src/main/java/com/github/onedirection/geolocation/geocoding/NominatimGeocoding.java
Ef55: I'm gonna try (and probably fail to) make it quick:

1. getBestNamedCoordinates was first implemented;
2. getNamedCoordinates was later implemented; 
3. getBestNamedCoordinates could now be implemented in the interface using it;
4. While implementing the cache, I realized the previous point which would reduce the amount of code needed for every new geocoding service (including the cache), so here you are.

TheImpressiveDonut: Okya nice !

TheImpressiveDonut (approved):

PR #143, opened 2021-04-26
Add a few features needed by the CachedDatabase to the Cache

Sosol02:
Still needs a few tests to be added before merging, but you can still review the code.

### DIFF
--- a/app/src/main/java/com/github/onedirection/utils/Cache.java
+++ b/app/src/main/java/com/github/onedirection/utils/Cache.java
@@ -14,14 +14,14 @@ import java.util.function.Function;
  * get(k) is O(backing) if k not computed yet, O(1) otherwise.
  * Note that the cache does not keep null values.
  *
- * getFunction is the function called to get elements,
- * setFunction is called when setting elements through the cache.
+ * getFunction is the default function called to get elements,
+ * setFunction is the default function called when setting elements through the cache.
  * It is imperative that setFunction(k, v); getFunction(k) == v
  * since in set(), the cache can remember (k, v) without calling
  * getFunction afterwards.
  *
- * If not specified, the default setFunction will throw a RuntimeException
- * and make the cache read-only.
+ * If not specified, the default get/setFunction will throw a RuntimeException
+ * and thus you will have to use the overloads of get and set that can accept a function.
  *
  * @param <K> The type of the keys
  * @param <V> The type of the values
@@ -83,20 +83,25 @@ public class Cache<K, V> {
     
     public void invalidate(K key) {
         map.remove(key);
-        // Do we actually want to suffer a fat O(n) cost here? maxHistory can be very large.
-        // We could not remove the key from the history, and when evicting keys, check if the
-        // removal failed (key was invalidated) and move on.
-        // Problems could happen because in this case, the number of useful history slots
-        // can only be at max maxHistory - #invalidate, which may be bad (force a lot of
-        // undue evictions).
-
-        //history.remove(key);
     }
 
+    /**
+     * Same as get(k, f) but using the default getFunction.
+     * @param key
+     * @return
+     */
     public V get(K key) {
         return get(key, getFunction);
     }
 
+    /**
+     * Get the value at key and use the function f to get it if it wasn't already in the cache.
+     * If the value corresponding to key is null, it is not stored in the cache.
+     * Complexity O(F) where F is the complexity of f.
+     * @param key
+     * @param f The function called if key is not in the cache.
+     * @return The value corresponding to key.
+     */
     public V get(K key, Function<? super K, ? extends V> f) {
         if (map.containsKey(key)) {
             return map.get(key);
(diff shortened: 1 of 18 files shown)


app/src/main/java/com/github/onedirection/utils/Cache.java
```diff
@@ -57,11 +57,19 @@ public Cache(
...
+        this(getFunction, (k, v) -> { throw new RuntimeException("Unreachable, set cache, k: " + k + ", v: " + v); }, maxHistory);
     }
 
     public Cache(Function<? super K, ? extends V> getFunction) {
-        this(getFunction, (k, v) -> { throw new RuntimeException("Read only cache."); }, MAX_HISTORY_DEFAULT);
+        this(getFunction, (k, v) -> { throw new RuntimeException("Unreachable, set cache, k: " + k + ", v: " + v); }, MAX_HISTORY_DEFAULT);
+    }
+
+    public Cache(int maxHistory) {
+        this(k -> { throw new RuntimeException("Unreachable, get cache, k: " + k); }, maxHistory);
```
Flechman, line 68: Should probably check that maxHistory is greater-eq than 0

app/src/main/java/com/github/onedirection/utils/Cache.java
```diff
@@ -57,11 +57,19 @@ public Cache(
     }
 
     public Cache(Function<? super K, ? extends V> getFunction, int maxHistory) {
-        this(getFunction, (k, v) -> { throw new RuntimeException("Read only cache."); }, maxHistory);
+        this(getFunction, (k, v) -> { throw new RuntimeException("Unreachable, set cache, k: " + k + ", v: " + v); }, maxHistory);
```
Ef55, line 60: Why "unreachable" ? The previous message seemed clearer.

app/src/main/java/com/github/onedirection/utils/Cache.java
Ef55: No need; the called constructor already does this checked.

Sosol02: It is already checked in the root constructor.

app/src/main/java/com/github/onedirection/utils/Cache.java
```diff
@@ -72,13 +80,29 @@ public void invalidate() {
...
+    public void invalidate(K key) {
+        map.remove(key);
+        // Do we actually want to suffer a fat O(n) cost here? maxHistory can be very large.
+        // We could not remove the key from the history, and when evicting keys, check if the
+        // removal failed (key was invalidated) and move on.
+        // Problems could happen because in this case, the number of useful history slots
+        // can only be at max maxHistory - #invalidate, which may be bad (force a lot of
+        // undue evictions).
+
+        //history.remove(key);
```
Ef55, line 93: I think this should be enabled:
- If we have performance issues, we can always turn it back off;
- Not having it may lead (at least is seems so to me) to unexpected behaviors, which might be hard to track.

Eventually, there could be a boolean argument to specify the behavior.

Sosol02:
atlas-coverage

Ef55 (changes requested):
I think that you should at least add documentation to some of the most "non-trivial" (e.g. set/get functions) behaviors.
(+ the other points need to be addressed)

app/src/main/java/com/github/onedirection/utils/Cache.java
Sosol02: I actually meant to remove that comment:
- The performance issues are not only issues but completely unacceptable since a going from O(1) to O(n) is simply not acceptable and in some way may defeat the purpose of having a cache.
- Using a boolean parameter seems to be a bad decision, it would mean leaking internal workings of the cache and make it more cubersome to implement the 2 alternatives at the same time while making sure no subtle bug happens.
- The behavior that you say may be "bad" or unexpected is very unlikely and the its only effect is to invalidate older entries a bit sooner than usual iff the cache is already full, and the effect doesnt last (once maxHistory elements are added no problem can ever ensue).
I will remove the comment and move my though process to the javadoc or something.

app/src/main/java/com/github/onedirection/utils/Cache.java
Sosol02: Yeah, the meaning of these functions has changed, will document.
The get() and set() functions now have overloads to allow passing in a custom get/setFunction, which makes the ones passed by constructor the default ones.

Ef55: I stand by my comment; unreachable is seemingly not the word you are looking for, as if this is printed, this was clearly reached.

Sosol02: It is written unreachable in the sense that this code should never have been called: if no default function is provided, it is an error to never provide a getFunction to get(k, f) and same for setFunction. "Unreachable" was more in the sense of "Should be unreachable but you reached it so you made a bad".
Also, 'Read only cache' was plain wrong since it is an intended use case to make caches with no default function and to provide one through get() and set() everytime, like in CachedDatabase, so you would end up with a cache that is not read only but with an error that would say so.
Maybe i should change it to "Should be unreachable" ? Or "No default get/setFunction provided" ?

Ef55: I'm with you on the fact that the old message didn't match the current behavior.
Yeah, I think you should change it to something like that.

Ef55 (approved):

Flechman (approved):
Clean mechanism for the cache. I guess you could later do a new PR to finish the  documentation you started, it would make the code cleaner  and would further help in understanding how the cache works.

PR #145, opened 2021-04-27
Loading screen

justagoodguy-omega:


app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CustomCalendarView.java
```diff
@@ -41,13 +42,13 @@
     private final SimpleDateFormat monthFormat = new SimpleDateFormat(("MMM"), Locale.ENGLISH);
     private final List<Date> dates = new ArrayList<>();
     private List<Event> eventsList = new ArrayList<>();
+    private final Activity parentActivity = (Activity) this.getContext();
```
Ef55, line 45: This doesn't look like a good idea.

app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CustomCalendarView.java
```diff
@@ -166,16 +164,28 @@ private int dateOfMonthAtPosition(int positionInGrid) {
     }
 
     private void callEventCreator(int year, int month, int day) {
-        Intent intent = new Intent(parentFragment.getActivity(), EventCreator.class);
+        Intent intent = new Intent(parentActivity, EventCreator.class);
```
Ef55, line 167: You could simply use a [context](https://developer.android.com/reference/android/content/Intent?hl=en#Intent(android.content.Context,%20java.lang.Class%3C?%3E)), which would avoid the ugly cast mentioned before.

app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CustomCalendarView.java
```diff
@@ -166,16 +164,28 @@ private int dateOfMonthAtPosition(int positionInGrid) {
...
         EventCreator.putDateExtra(intent, LocalDate.of(year, month, day));
-        parentFragment.startActivity(intent);
+        parentActivity.startActivity(intent);
```
Ef55, line 169: Same here, the context should be enough.

Ef55 (changes requested):
Happy to see that you learn from your past mistakes e13babf :)

Sosol02 (approved):
Looks okay.

app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CustomCalendarView.java
```diff
@@ -166,16 +163,28 @@ private int dateOfMonthAtPosition(int positionInGrid) {
...
         //@TODO when the events list is implemented;
     }
 
-    private int getMonthNumber(Calendar cal){
+    private int getMonthNumber(Calendar cal) {
         return cal.get(Calendar.MONTH) + 1;
     }
+
+    private LoadingDialog startLoadingAnimation() {
+        LoadingDialog loadingDialog = new LoadingDialog((Activity) this.getContext());
```
Ef55, line 180: Moving it doesn't make it better :)

Ef55 (changes requested):
You also included some firebase-emulator configuration files, which should be removed.

app/src/main/java/com/github/onedirection/utils/LoadingDialog.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public LoadingDialog(Activity hostActivity) {
+        activity = hostActivity;
+    }
+
+    @SuppressLint("InflateParams")
+    public void startLoadingAnimation() {
+
+        AlertDialog.Builder builder = new AlertDialog.Builder(activity);
+        LayoutInflater inflater = activity.getLayoutInflater();
```
Ef55, line 22: (Part of my previous review; dunno why github didn't include it)
You might use 
```suggestion
        LayoutInflater inflater = (LayoutInflater)context.getSystemService(Context.LAYOUT_INFLATER_SERVICE);
```
which would solve the aforementioned casting issue, as you did at [CustomCalendarView.java:111](https://github.com/Sosol02/sdp/blob/loadingScreen/app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CustomCalendarView.java#L111).

app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CustomCalendarView.java
justagoodguy-omega: Well here it literally asks for an activity :'(

Ef55 (approved):

PR #146, opened 2021-04-27
Recurring event refactor

Flechman:
- No more pointers to prev event and next event in the Recurrence class (speeds up the queries a lot)
- Got rid of the following invariant : a recurrence series has more than 1 element (events). Now, this is authorized, as it doesn't bother anything and it isn't necessary.
- Added a method to modify an event. This method makes any modification on an event, except if it is recurrent (in that case modifying the recurrence series id or the period of the recurrence is forbidden, whereas modifying the end time of the recurrence is accepted and supported).

### DIFF
--- a/app/src/main/java/com/github/onedirection/EventQueries.java
+++ b/app/src/main/java/com/github/onedirection/EventQueries.java
@@ -131,6 +131,7 @@ public class EventQueries {
      * given in the Recurrence attribute of 'event', and the time between each recurrence of the event is based on the period given in the Recurrence attribute of 'event'.
      * Ex use : addRecurringEvent(event("tennis", startTime: "Jan 1 2021).Recurrence(period:"weekly", endTime: " Jan 1 2022")) means the event "tennis" will occur every week
      * starting at the begining of 2021 and will end at the begining of 2022
+     * Note : the new recurrence series should have as id the given 'event' id (this event is called the root of the recurrence series).
      * @param event (Event) : the given event representing the recurrence series
      * @return (CompletableFuture<Integer>) : If the recurrence series has been successfully created and stored into the database it returns the number of stored events, else it returns 0
      */
@@ -140,6 +141,9 @@ public class EventQueries {
         }
         if(event.getStartTime().toEpochSecond() > event.getRecurrence().get().getEndTime().toEpochSecond()) {
             throw new IllegalArgumentException("The end time is invalid");
+        }
+        if(!event.getRecurrence().get().getGroupId().equals(event.getId())) {
+            throw new IllegalArgumentException("The given recurrence series id does not match the root event id");
         }
         List<Event> eventsToStore = new ArrayList<Event>();
         Recurrence newEventRecurrence = event.getRecurrence().get();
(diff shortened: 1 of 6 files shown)


Ef55 (commented):
Code-wise, I don't really have anything to say.
I just have a few questions:

- Is there a particular reason why the period cannot be changed ?
- What happens if the recurrence is removed ? I saw code handling all other cases, but I'm not sure for this one.

Also, if you ever want to modularize this, a possibility would be to move the logic instantiating "simple" events from a recurring one into another file.

Flechman:
> Code-wise, I don't really have anything to say.
> I just have a few questions:
> 
> * Is there a particular reason why the period cannot be changed ?
> * What happens if the recurrence is removed ? I saw code handling all other cases, but I'm not sure for this one.
> 
> Also, if you ever want to modularize this, a possibility would be to move the logic instantiating "simple" events from a recurring one into another file.

About the last point : I'm not sure it's the best thing to do since many methods don't make the difference between recurring and non-recurring events.
- The period cannot be changed because of the freedom we give to modifying an event that is recurrent : If for example we delete an event in a recurrence series or we change just once the time of the event, recovering the information when we want to change the period of the recurrence is extremely costly (in some cases it can even make no sense because the times of each event of the recurrence series have changed). Instead of "reinitializing" each  event of the recurrence series when we change the period, I have decided to forbid the possibility of changing the  period.
- Via the method 'modifyEvent' one cannot remove a full recurrence series. You will have to call specifically the method that does this. 
- If you meant removing the recurrence object from the event, then this is supported. It will just mean that the event is unlinked to the recurrence series.

Ef55:
I'm still a bit puzzled about the restriction on the period, as I'm not sure this makes sense from a user stand-point.
Also aren't recurring events stored as one single entity in the DB ? If so, why would it be costly to update the period ?

Regarding deletion, I meant "remove the recurrence object from the event".

Also, now that you dropped the invariant regarding the size of the recurrence, would it be possible to make so that the recurrence Id is the Event id of the "root event" (the first one) ? I'm proposing this to simplify Id generation.

Flechman:
> I'm still a bit puzzled about the restriction on the period, as I'm not sure this makes sense from a user stand-point.
> Also aren't recurring events stored as one single entity in the DB ? If so, why would it be costly to update the period ?
> 
> Regarding deletion, I meant "remove the recurrence object from the event".
> 
> Also, now that you dropped the invariant regarding the size of the recurrence, would it be possible to make so that the recurrence Id is the Event id of the "root event" (the first one) ? I'm proposing this to simplify Id generation.

The restriction on the period enables to prevent from having conflicts with  the changes made on certain events belonging to the recurrence series. So it does make sens from a user point of view. What makes less sense is the fact that the user wants to change the recurrence period for all the events in the recurrence series, he would be better off deleting the current recurrence series and then starting over and create a new recurrence series.

Each event (recurring or not) is stored as one entity in the DB. When I say it's costly I mean trying to keep the changes made on each event of the  recurrence series. I can give you an example whenever we have the time to make a call.

The last point is a good idea, I don't see any objections regarding this change.

app/src/main/java/com/github/onedirection/EventQueries.java
```diff
@@ -131,6 +131,7 @@ public EventQueries(Database db) {
      * given in the Recurrence attribute of 'event', and the time between each recurrence of the event is based on the period given in the Recurrence attribute of 'event'.
      * Ex use : addRecurringEvent(event("tennis", startTime: "Jan 1 2021).Recurrence(period:"weekly", endTime: " Jan 1 2022")) means the event "tennis" will occur every week
      * starting at the begining of 2021 and will end at the begining of 2022
+     * Note : the new recurrence series should have as id the given 'event' id (this event is called the root of the recurrence series).
```
Ef55, line 134: [Nitpick] I believe there is a typo.

Ef55 (approved):

Sosol02 (approved):
i trust you
i am tired
tuez les otages

PR #147, opened 2021-04-27
Add a temporary fix to broken mapbox test

Sosol02:


TheImpressiveDonut (approved):

app/src/androidTest/java/com/github/onedirection/map/MapFragmentTest.java
```diff
@@ -143,6 +143,8 @@ public void testClickOnMapPutAMarkerOnTheMap() {
 
     @Test
     public void testAddEventPutsMarkerOnMap() {
+        // Wait acton to make getMarkerSymbolManager work.
```
Flechman, line 146: [detail] Not necessary but could specify that it is a race condition

Flechman (approved):
Thanks for correcting this, test should now be deterministic

PR #149, opened 2021-04-29
Sol cached database

Sosol02:


app/src/androidTest/java/com/github/onedirection/database/Item.java
```diff
@@ -12,7 +12,7 @@
     private final String s;
 
     // used by storer
-    private Item(Id id, int number, String s) {
+    Item(Id id, int number, String s) {
```
Flechman, line 15: Shouldn't you just put public there in that case ?

Flechman (commented):
The changes are almost nothing but refactoring so i guess things are kept clean.
Why did you create a MockDatabase class ?

Sosol02:
The mocking database is to avoid creating a very big class in the last test of CachedDatabaseTests, and to allow code reuse in the future.

Flechman (approved):
Since the PR concerns mainly refactoring, there are no big concerns. Approved and thumbs up.

hugolan (approved):

PR #150, opened 2021-04-29
Event creator callback

Ef55:
Edit EventCreator's callback to create the event, in order to ease testing.
Rewrite all tests to be more stringent.

Also change default callback to use the newest EventQueries methods.


app/src/main/java/com/github/onedirection/events/EventCreator.java
```diff
@@ -111,22 +101,33 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            this.model.init(getDateExtra(intent), EventCreator::putEventToDatabase);
+        } else {
+            this.model.init(EventCreator::putEventToDatabase);
         }
-        else if(hasDateExtra(intent)){
-            this.model.init(getDateExtra(intent));
+    }
+
+    private static void putEventToDatabase(Event event, boolean edited) {
+        Log.d(LOGCAT_TAG, event.toString());
```
Flechman, line 122: If this line is only here for debugging the tests i guess you can get rid of it

app/src/main/java/com/github/onedirection/events/EventCreatorMainFragment.java
```diff
@@ -1,222 +1 @@
...
                time.getValue().getDayOfMonth()
        );
        datePicker.show();
    }

    private Event generateEvent() {
        return model.generateEvent();
    }

}
```
Flechman, line 1: Hmmmmm very weird way of writing code. Should probably be fixed.

Flechman (changes requested):
I think it is necessary that you check the comments i put.

app/src/main/java/com/github/onedirection/events/EventCreator.java
Ef55: Well, it's Logcat, so I think this can be left here (but I should verify this information).

app/src/main/java/com/github/onedirection/events/EventCreatorMainFragment.java
Ef55: Weird indeed; gonna fix this.

justagoodguy-omega (changes requested):
Just change the weird scrambled file, otherwise it looks good

Ef55:
Regarding the weird `^M`; I suspect that this is a GitHub issue, as the file looks fine in local (even after pulling/formating/pushing back).

justagoodguy-omega (approved):

app/src/main/java/com/github/onedirection/events/EventCreator.java
Flechman: After looking on internet, it seems to be fine to keep log messages during development cycle. At release time some dev tools exist to get rid of log messages before building the release version. So it's fine for the moment

Flechman (approved):
Feedback has been addressed, seems all good for me.

PR #166, opened 2021-05-04
Implement cache dumping

Ef55:


app/src/test/java/com/github/onedirection/utils/CacheTest.java
```diff
@@ -198,4 +220,60 @@ public void writeAllocateWorks() {
...
+
+        Cache<Integer, Integer> ref = new Cache<>(get, 2);
+        for (int i = 0; i <= 9; ++i) {
+            ref.get(i);
+        }
+
+
+        ByteArrayOutputStream output = new ByteArrayOutputStream();
+        boolean dumpSuccess = ref.dumpToStream(output, i -> i, i -> i);
+        assertThat(dumpSuccess, is(true));
```
hugolan, line 241: what happens if dumpsuccess is false?

hugolan (approved):

app/src/test/java/com/github/onedirection/utils/CacheTest.java
Ef55: Since I used a `ByteArrayOutputStream` in this test, `dumpSuccess` should never be false, except if the implementation is buggy, in which case an assertion error will be thrown by line 241 (which is what we want).

justagoodguy-omega (approved):
From my limited knowledge of Chinese this looks good

hugolan (approved):

app/src/main/java/com/github/onedirection/utils/Cache.java
```diff
@@ -33,20 +43,42 @@
...
         // avoid edge case where functions expect a non empty cache
         if (maxHistory < 1) throw new IllegalArgumentException();
         this.getFunction = Objects.requireNonNull(getFunction);
         this.setFunction = Objects.requireNonNull(setFunction);
         this.maxHistory = maxHistory;
-        this.history = new LinkedBlockingQueue<>(maxHistory);
+        this.map = map;
+        this.history = history;
+
+        // Note: I did not a find a way to verify that the capacity of the queue
```
Flechman, line 72: In the documentation of LinkedBlockingQueue it is said 
```
LinkedBlockingQueue(int capacity)
Creates a LinkedBlockingQueue with the given (fixed) capacity.
```

So I guess you  don't have to think about checking.

Flechman (approved):
My comments are not decisive changes, they are just there for my better understanding.

app/src/main/java/com/github/onedirection/utils/Cache.java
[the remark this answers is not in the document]
Ef55: What I meant by this comment is that I have no way of ensuring that
```
history.capacity == maxHistory
```
holds. This is not a huge issue as this constructor is private, so we can simply assume that the caller has the responsibility for this, but it annoyed me (and I love to complain about this kind of things in my comments).

app/src/main/java/com/github/onedirection/utils/Pair.java
```diff
@@ -11,6 +20,10 @@ public Pair(S first, T second){
         this.second = second;
     }
 
+    public static<S, T> Pair<S, T> of(S first, T second){
+        return new Pair(first, second);
```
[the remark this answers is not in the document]
Ef55: Fair point. At no point we check that the arguments are not null.
In this particular case (where the class is used more like a struct than a real class), I prefer leaving the user free of having null attributes if he wants so. 
Of course this point is debatable, but here I don't think it's worth our time.

PR #168, opened 2021-05-05
Event creator fixes

Ef55:
Don't let the number of files/lines scare you:

- Many files were moves/renamed
- MainFragment got an encoding fix which make git mark the whole file as edited (nothing really changed)


app/src/main/java/com/github/onedirection/events/ui/MainFragment.java
```diff
@@ -0,0 +1,220 @@
+package com.github.onedirection.events.ui;
```
Flechman, line 1: [Not related to the code line] Ahhh  it was the file that was weirdly encoded with the M^

Flechman (approved):
File encoding on MainFragment class has been corrected this is great!

app/src/main/java/com/github/onedirection/events/ui/MainFragment.java
Ef55: Yep. I found the reason [here](https://stackoverflow.com/questions/1822849/what-are-these-ms-that-keep-showing-up-in-my-files-in-emacs).
TLDR; Unix uses LF for new lines, windows uses CRLF. CRLF are normally always converted into LF by git.
For some reason, this wasn't the case for this file, and the CR were displayed as ^M.
I fixed it using VScode, which allow to convert new lines encoding.

Flechman: Ok good to know for later development hahaha

Sosol02 (approved):
Mostly refactoring, looks ok to me

PR #170, opened 2021-05-05
Add utility to use Cache-class with Android-cache

Ef55:


app/src/main/java/com/github/onedirection/utils/Cache.java
```diff
@@ -36,7 +36,7 @@
  * @param <K> The type of the keys
  * @param <V> The type of the values
  */
-public class Cache<K, V> {
+public final class Cache<K, V> {
```
Sosol02, line 39: Although we don't have a use case for a subclass right now, I don't see why we want to deny inheritance for that specifically. For instance, it's the kind of class that could be easily mocked by subclassing.

Sosol02 (approved):
Looks good, clean code

app/src/main/java/com/github/onedirection/utils/Cache.java
Ef55: I tend to apply the principle that classes are either abstract or final, so that's the reason why (but I don't really care, we can change it back if need be).

hugolan (approved):

PR #174, opened 2021-05-06
Changes taking into account the code review by the TAs

Flechman:


app/src/main/java/com/github/onedirection/database/queries/EventQueries.java
```diff
@@ -1,28 +1,25 @@
...
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * A class used to query the events in a database based on different operations
  */
 public class EventQueries {
 
+    @SuppressWarnings("FieldMayBeFinal")
```
justagoodguy-omega, line 22: Is this really the best fix?

app/src/main/java/com/github/onedirection/database/store/EventStorer.java
```diff
@@ -28,6 +32,9 @@
     public static final String KEY_RECURR_END_TIME = "recurrEndTime";
     public static final String KEY_RECURR_PERIOD = "recurrPeriod";
 
+    private EventStorer() {
+    }
```
justagoodguy-omega, line 36: I think you can remove this since the default constructor already does it

justagoodguy-omega (changes requested):

app/src/main/java/com/github/onedirection/database/store/EventStorer.java
Flechman: It's the whole point of private constructors. 
The default constructor does this only if there are no public + private constructors. With the empty private constructor, I prevent any default constructor and thus any object creation out of this class.
PS : many other classes in the project have this construct

app/src/main/java/com/github/onedirection/database/queries/EventQueries.java
Flechman: The problem is the warning, that suggests to put the attribute 'db' to final.
THEORETICALLY (we'll never do it in practice, but it's to stay consistent) we may change the database (ex depending on the device) and thus the 'db' attribute.

app/src/main/java/com/github/onedirection/event/EventCreatorMainFragment.java
```diff
@@ -0,0 +1 @@
...
                time.getValue().getDayOfMonth()
        );
        datePicker.show();
    }

    private Event generateEvent() {
        return model.generateEvent();
    }

}
```
hugolan, line 1: I am the only one to see buggy caracter on this file?

app/src/main/java/com/github/onedirection/navigation/fragment/sign/SignViewModel.java
```diff
@@ -31,7 +30,7 @@ public void sign(String username, String password, boolean register) {
                     .exceptionally(error -> {userResult.setValue(null); return null; });
         } else {
             auth.loginUser(username, password).thenAccept(userResult::setValue)
-                    .exceptionally(error -> {userResult.setValue(null); return null; });;
+                    .exceptionally(error -> {userResult.setValue(null); return null; });
```
hugolan, line 33: nice fix

hugolan (approved):

Flechman:
The PR has been closed because some changes resulted in some tests not working and I couldn't find the origin of this.

PR #179, opened 2021-05-06
Implement sync from db to map and fix tests

Sosol02:


app/src/main/java/com/github/onedirection/utils/Monads.java
```diff
@@ -25,6 +25,11 @@ private Monads() {
         return m.thenCompose(o -> toFuture(o));
     }
 
+    // ahh monads.. if only
```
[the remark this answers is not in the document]
Sosol02: ur right forgot to remove

Flechman (approved):
Code added seems fine by me.

app/src/main/java/com/github/onedirection/navigation/fragment/map/MarkerSymbolManager.java
```diff
@@ -146,4 +149,26 @@ private void initializeSymbolImage(Context context, Style style) {
     public Map<Symbol, Event> getEventMap() {
         return Collections.unmodifiableMap(eventMap);
     }
+
+    public CompletableFuture<Void> syncEventsWithDb() {
+        Log.d(LOG_TAG, "syncEventsWithDb");
```
TheImpressiveDonut, line 154: Nice, does the db works with the cache? Because I am not really sure of how exactly mapfragment keep state.

TheImpressiveDonut (approved):

app/src/main/java/com/github/onedirection/navigation/fragment/map/MarkerSymbolManager.java
Sosol02: Yes it uses the Database.getInstance db, which since last week is the cached one by default.

PR #180, opened 2021-05-06
A few changes taking into account the code review by our coaches

Flechman:


app/src/main/java/com/github/onedirection/database/queries/EventQueries.java
```diff
@@ -1,28 +1,25 @@
...
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * A class used to query the events in a database based on different operations
  */
 public class EventQueries {
 
+    @SuppressWarnings("FieldMayBeFinal")
```
hugolan, line 22: is this the best fix?

hugolan (approved):
nice pr fix

app/src/main/java/com/github/onedirection/database/queries/EventQueries.java
Flechman: The problem is the warning, that suggests to put the attribute 'db' to final.
We could change the database (ex depending on the device) and thus the 'db' attribute so putting it to final is not consistent with the possibilities.

PR #197, opened 2021-05-09
Fix NominatimGeocoding to follow guidelines

Ef55:
Add mechanism to space requests by at least 1sec.
Add user-agent to requests.


Sosol02:
How is the limit implemented? Does the returned future throw an exception, is it put into a queue? Just curious

app/src/main/java/com/github/onedirection/geolocation/geocoding/NominatimGeocoding.java
```diff
@@ -109,27 +114,50 @@ private static String generateReverseRequestURL(Coordinates coordinates){
...
+        S request = valReqPair.second;
+
+        result = result.whenComplete((t, throwable) -> {
+            if (throwable != null) {
+                request.cancel();
+            }
+        });
+
+
+        long delay = Math.min(0, MIN_DELAY_MS - (SystemClock.uptimeMillis() - this.lastRequest));
```
Flechman, line 137: Is it min or is it max ? I suppose that a delay can be positive (and is usually positive) so having either 0 or a negative value seems incorrect ?

Flechman (changes requested):

app/src/main/java/com/github/onedirection/geolocation/geocoding/NominatimGeocoding.java
Ef55: Nice catch ! This should be fixed now.
(`postDelayed` seemingly does this `max(0, ...)` operation itself).

Ef55:
> How is the limit implemented? Does the returned future throw an exception, is it put into a queue? Just curious

In all honesty, I forgot to test what is the behavior of Nominatim in this case; It would most likely either return 50 results (so everything is fine except that you end up with less results than expected) or reject the request (in which case the future completes exceptionally).

Flechman (approved):

app/src/main/java/com/github/onedirection/geolocation/geocoding/NominatimGeocoding.java
```diff
@@ -109,27 +115,51 @@ private static String generateReverseRequestURL(Coordinates coordinates){
...
+        S request = valReqPair.second;
+
+        result = result.whenComplete((t, throwable) -> {
+            if (throwable != null) {
+                request.cancel();
+            }
+        });
+
+
+        long delay = Math.max(0, MIN_DELAY_MS - (SystemClock.uptimeMillis() - this.lastRequest));
```
Sosol02, line 138: You should store SystemClock.uptimeMillis in a variable and use it for here and line 140: the times may be different and the clock may not be monotonic.
I think it's not very important here but it's a good habit to have.

Sosol02 (changes requested):

app/src/main/java/com/github/onedirection/geolocation/geocoding/NominatimGeocoding.java
Ef55: In this particular case, this is intentional :)
The reason being that I want `lastRequest` to be after the handler was posted, in case there is some delay between the moment between the delay computation and the request posting.

Sosol02: oki

Sosol02 (approved):

PR #198, opened 2021-05-10
Fix test mapquest routeservice

TheImpressiveDonut:
Create a route service mock so we can test the navigation manager and routes manager

### DIFF
--- a/app/src/androidTest/java/com/github/onedirection/map/RouteServiceMock.java
+++ b/app/src/androidTest/java/com/github/onedirection/map/RouteServiceMock.java
@@ -2,7 +2,6 @@ package com.github.onedirection.map;
 
 import androidx.annotation.NonNull;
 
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/com/github/onedirection/map/RouteServiceMock.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ *    Route service that return route between those points
+ *    LatLng(40.7326808, -73.9843407);
+ *    LatLng(40.7326808, -73.9843407);
+ */
+
+public class RouteServiceMock implements RouteService {
```
Flechman, line 35: Could add a small line of comment to explain the role of the class

Flechman (approved):
There's not much to say, the class seems to correctly describe a route. Hope it's helpful for the tests

PR #199, opened 2021-05-10
Refactor default databases and add ObservableDatabase.

Sosol02:


Sosol02:
Still need to write tests but this pr is mostly refactoring and encapsulating so it should be allright.

app/src/main/java/com/github/onedirection/database/DefaultDatabase.java
```diff
@@ -0,0 +1,14 @@
+package com.github.onedirection.database;
+
+import androidx.annotation.VisibleForTesting;
+
+public class DefaultDatabase {
```
Flechman, line 5: What does this class do ? A small line to describe it could be nice

app/src/main/java/com/github/onedirection/database/ObservableDatabase.java
```diff
@@ -0,0 +1,265 @@
...
+
+import static com.github.onedirection.utils.ObserverPattern.*;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+
+public class ObservableDatabase implements Database, Observable<ObservableDatabase.Action> {
```
Flechman, line 15: Same small comment here

app/src/main/java/com/github/onedirection/database/ObservableDatabase.java
```diff
@@ -0,0 +1,265 @@
...
+    ObservableDatabase(Database innerDb) {
+        this.innerDb = Objects.requireNonNull(innerDb);
+    }
+
+    private final ArrayList<Observer<Action>> observers = new ArrayList<>();
+    private final Database innerDb;
+
+    @Override
+    public boolean addObserver(Observer<Action> observer) {
+        observers.add(Objects.requireNonNull(observer));
```
Flechman, line 26: Could directly `return observers.add(Objects.requireNonNull(observer));` instead of adding, and then just returning true.

app/src/main/java/com/github/onedirection/database/ObservableDatabase.java
```diff
@@ -0,0 +1,265 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean removeObserver(Observer<Action> observer) {
+        observers.add(Objects.requireNonNull(observer));
```
Flechman, line 32: Same here

app/src/main/java/com/github/onedirection/navigation/NavigationActivity.java
```diff
@@ -66,7 +66,7 @@ public boolean onMenuItemClick(MenuItem menuItem) {
                 startActivity(intent);
                 return false;
             }
-        });
+        }); // */
```
Flechman, line 69: Can remove this I suppose (thankssss)

Flechman (changes requested):
You have mastered the CompletableFuture manipulation

app/src/main/java/com/github/onedirection/navigation/NavigationActivity.java
Sosol02: ok

app/src/main/java/com/github/onedirection/database/ObservableDatabase.java
Sosol02: Allows registering callbacks on database actions, will document.

app/src/main/java/com/github/onedirection/database/ObservableDatabase.java
Sosol02: only api i know where add returns a boolean xd, will do

app/src/main/java/com/github/onedirection/database/ObservableDatabase.java
Sosol02: ssame

Flechman (approved):
Changes seem fine! AOK for me

justagoodguy-omega (approved):
Seems good to me

PR #203, opened 2021-05-11
Minor calendar fix

justagoodguy-omega:


app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CalendarGridAdapter.java
```diff
@@ -67,16 +71,24 @@ public View getView(int position, @Nullable View convertView, @NonNull ViewGroup
                         && displayYear == eventCalendar.get(Calendar.YEAR)) {
                     nbOfEventsInDay++;
                     if (nbOfEventsInDay == 1) {
-                        EventNumber.setText(nbOfEventsInDay + " Event");
+                        EventNumber.setText(String.format("%s Event", nbOfEventsInDay));
```
Ef55, line 74: Format is not really needed here; you know that 
`nbOfEventsInDay == 1`.
Also, as requested in the coaches' review, such strings should be in strings.xml.

app/src/main/java/com/github/onedirection/navigation/fragment/calendar/DayEventsListView.java
```diff
@@ -86,7 +88,6 @@ private AlertDialog setupDialog(List<Event> events) {
         } else {
```
Sosol02, line 88: I think it would be cleaner if the function first checked for events.size() != 0, then inside it create and configure the alertDialog, instead of creating it and later dissmissing it.
Also (nitpick) you may want to Objects.requireNonNull(events) to avoid the NullPointerException.

Sosol02 (changes requested):
nice dock

Sosol02 (approved):

app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CalendarGridAdapter.java
```diff
@@ -67,16 +71,24 @@ public View getView(int position, @Nullable View convertView, @NonNull ViewGroup
                         && displayYear == eventCalendar.get(Calendar.YEAR)) {
                     nbOfEventsInDay++;
                     if (nbOfEventsInDay == 1) {
-                        EventNumber.setText(nbOfEventsInDay + " Event");
+                        EventNumber.setText(R.string.one_event_cal_display);
                     } else {
-                        EventNumber.setText(nbOfEventsInDay + " Events");
+                        EventNumber.setText(String.format("%s Events", nbOfEventsInDay));
```
Ef55, line 76: This should also be in the strings.xml (sorry for not highlighting it earlier).

justagoodguy-omega: I looked at it and I didn't see how to put it in string.xml since it needs nbOfEventsInDay as argument, are you sure it's possible?

Ef55: You should move the format string (`"%s Events"`) in _string.xml_, and do `String.format(R.string.event_count_format, nbOfEventsInDay)` on it.

justagoodguy-omega: Yeah I'd tried to do that but it doesn't compile because String.format doesn't like R.string.event_count_format as argument

Ef55: From [here](http://www.envyandroid.com/android-string-xml-resource-formatting/), and with a bit of generalization:
```
String.format(view.getContext().getResources().getString(R.string.event_count_format), nbOfEventsInDay)
```
should work.

justagoodguy-omega: cOde ReAdaBiLiTy

Ef55: Fair point 🙃

Ef55 (approved):

PR #210, opened 2021-05-13
Change small text specifying number of events in calendar view + other minor UI changes

Flechman:


hugolan (approved):

app/src/main/java/com/github/onedirection/navigation/fragment/calendar/CalendarGridAdapter.java
```diff
@@ -59,24 +61,25 @@ public View getView(int position, @Nullable View convertView, @NonNull ViewGroup
...
                     nbOfEventsInDay++;
-                    if (nbOfEventsInDay == 1) {
-                        EventNumber.setText(R.string.one_event_cal_display);
-                    } else {
-                        EventNumber.setText(String.format(view.getContext().getResources().getString(R.string.multiple_events_cal_display), nbOfEventsInDay));
-                    }
                 }
             }
+            if(nbOfEventsInDay > 0) {
+                eventNumber.setText(String.format(view.getContext().getResources().getString(R.string.events_call_display), nbOfEventsInDay));
```
Ef55, line 78: I think you could simply use `Integer.toString(nbOfEventsInDay)` instead of using a format string (and this would greatly simplify the call).

Ef55 (changes requested):

Ef55 (approved):

PR #224, opened 2021-05-14
Make day events list close automatically when there are no events

justagoodguy-omega:


Ef55 (changes requested):
This modification is welcome, but there is a slight issue:
if you press "View events" on a day without any events, nothing happens.

In my opinion, the button should be clearly disabled/gone in this case (ev. with a message like "No events to display").

justagoodguy-omega:
> 
> 
> This modification is welcome, but there is a slight issue:
> if you press "View events" on a day without any events, nothing happens.
> 
> In my opinion, the button should be clearly disabled/gone in this case (ev. with a message like "No events to display").

The thing it has been the case for a while now so we could change it but it isn't linked to this PR

Ef55:
> > 
> > 
> > This modification is welcome, but there is a slight issue:
> > if you press "View events" on a day without any events, nothing happens.
> > 
> > In my opinion, the button should be clearly disabled/gone in this case (ev. with a message like "No events to display").
> 
> The thing it has been the case for a while now so we could change it but it isn't linked to this PR

The thing is that, before, there just was an empty list, but now there is nothing (gives the sensation that the button doesn't work).
Anyway, we'll fix this in a subsequent PR.

Ef55 (approved):

justagoodguy-omega:
> 
> 
> > > This modification is welcome, but there is a slight issue:
> > > if you press "View events" on a day without any events, nothing happens.
> > > In my opinion, the button should be clearly disabled/gone in this case (ev. with a message like "No events to display").
> > 
> > 
> > The thing it has been the case for a while now so we could change it but it isn't linked to this PR
> 
> The thing is that, before, there just was an empty list, but now there is nothing (gives the sensation that the button doesn't work).
> Anyway, we'll fix this in a subsequent PR.

The thing is that nope, before it did the same thing, what you're thinking of is when the user deletes an event and there are no more remaining there used to be a blank screen which was indeed changed in this PR

Ef55:
Oh ok, sorry about that; I wrongfully assumed that it was previously possible to open the blank page.

TheImpressiveDonut (approved):
Nice code, not much changed so I have nothing to say !

hugolan (approved):

PR #225, opened 2021-05-15
Allow GeocodingCache dumping and integrate it in the app

Ef55:
The cache is automatically loaded from file on app startup.
The cache is dumped every minute to file.


app/src/main/java/com/github/onedirection/geolocation/geocoding/GeocodingCache.java
```diff
@@ -1,34 +1,66 @@
...
+    public GeocodingCache(GeocodingService underlying) {
+        this(
+                new Cache<>(
+                        key -> underlying.getNamedCoordinates(key.first, key.second)
+                ),
+                new Cache<>(
+                        underlying::getBestNamedCoordinates
+                )
         );
+        Objects.requireNonNull(underlying);
```
Flechman, line 63: Aren't you supposed to check this condition before anything you do in the  constructor ? I feel like it should be line 55.

Flechman (changes requested):
The code is very neat. I prefer that you review my single comment before I approve

app/src/main/java/com/github/onedirection/geolocation/geocoding/GeocodingCache.java
Ef55: Not possible: when you call a constructor (from another constructor), it must be the first thing you do.

Flechman: Ahhh yes I forgot the rule

Flechman: Isn't it unsafe to call getNamedCoordinates then ?

Ef55: The method is called inside a lambda; said lambda is not called in the constructor, only later.
As such, this call to `Objects.requireNonNull(...)` should be sufficient.

Flechman (approved):

app/src/main/java/com/github/onedirection/geolocation/geocoding/GeocodingCache.java
```diff
@@ -40,4 +72,33 @@ public GeocodingCache(GeocodingService underlying){
     public CompletableFuture<NamedCoordinates> getBestNamedCoordinates(Coordinates coordinates) {
         return coordinatesCache.get(coordinates);
     }
+
+    public boolean dumpToAndroidCache(Context ctx, String filename) {
+        return AndroidCache.dumpToAndroidCache(ctx, filename + "_1", nameCache, x -> x, GeocodingCache::dumpList)
+                && AndroidCache.dumpToAndroidCache(ctx, filename + "_2", coordinatesCache, x -> x, cf -> cf.getNow(null));
+    }
+
+    public static Optional<GeocodingCache> loadFromAndroidCache(Context ctx, String filename, GeocodingService underlying) {
```
TheImpressiveDonut, line 81: Since you tested if underlying is not null in the constructor, shouldn't you check it too in this method?

app/src/main/java/com/github/onedirection/geolocation/geocoding/GeocodingService.java
```diff
@@ -59,3 +74,51 @@
     CompletableFuture<NamedCoordinates> getBestNamedCoordinates(Coordinates coordinates);
 
 }
+
+class GeocodingServices {
```
TheImpressiveDonut, line 78: Just wondering why is there a class in an interface?

TheImpressiveDonut (changes requested):
Nice code ! Just a few minor things to see.

app/src/main/java/com/github/onedirection/utils/Cache.java
```diff
@@ -247,7 +250,12 @@ boolean dumpToStream(
     ){
         try(ObjectOutputStream outputStream = new ObjectOutputStream(target)) {
             outputStream.writeInt(maxHistory);
-            outputStream.writeObject(Monads.map(map, combine(keyMap, valMap)));
+            outputStream.writeObject(
+                    map.entrySet().stream()
+                            .map(e -> combine(keyMap, valMap).apply(e.getKey(), e.getValue()))
+                            .filter(e -> e.getKey() != null && e.getValue() != null)
+                            .collect(toMap(Map.Entry::getKey, Map.Entry::getValue))
```
hugolan, line 257: [
image
](url)

nice code!

app/src/main/java/com/github/onedirection/geolocation/geocoding/GeocodingCache.java
```diff
@@ -40,4 +72,33 @@ public GeocodingCache(GeocodingService underlying){
...
+        );
+        Optional<Cache<Coordinates, CompletableFuture<NamedCoordinates>>> coordinatesCache = AndroidCache.loadFromAndroidCache(
+                ctx,
+                filename + "_2",
+                underlying::getBestNamedCoordinates,
+                (Coordinates x) -> x,
+                (Function<NamedCoordinates, CompletableFuture<NamedCoordinates>>) CompletableFuture::completedFuture
+        );
+
+        if(nameCache.isPresent() && coordinatesCache.isPresent()){
```
hugolan, line 97: wouldn't it also work if only one of them is present in the android cache?

hugolan (approved):

app/src/main/AndroidManifest.xml
```diff
@@ -36,7 +36,7 @@
             android:name="androidx.startup.InitializationProvider"
             android:exported="false" >
             <meta-data
-                android:name="com.github.onedirection.startup.NotificationInitializer"
+                android:name="com.github.onedirection.startup.ApplicationInitializer"
```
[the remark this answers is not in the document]
Ef55: I renamed it because it is now also used to initialize the Geocoding default instance(s), so the previous name was a bit misleading.

app/src/main/java/com/github/onedirection/geolocation/geocoding/GeocodingService.java
Ef55: The class is not in (unless I messed up) in the interface but next to it. 
Java doesn't allow static attributes in interfaces, so I moved the code generating/storing the static instances into this package private class. The interface then retrieves the static fields and returns them as-if they were its own members.

app/src/main/java/com/github/onedirection/geolocation/geocoding/GeocodingCache.java
Ef55: Indeed; this case should never happen, so I didn't bother handling it in the most efficient way.
This could be an improvement but as said, it shouldn't happen, so I don't think it's worth the effort.

TheImpressiveDonut (approved):

PR #227, opened 2021-05-16
Sanitize user inputs in the Event Creator view 

Flechman:


app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
```diff
@@ -41,9 +46,11 @@
  */
 public class MainFragment extends Fragment {
 
+    @VisibleForTesting
+    final static int MAX_STRING_LENGTH = 100;
```
hugolan, line 50: isn't 100 too much?

hugolan (changes requested):

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
```diff
@@ -224,15 +232,45 @@ private Event generateEvent() {
...
-    private void addEventCallback(View v) {
-        model.incrementLoad();
-        requireActivity().findViewById(R.id.eventCreatorMainFragment).setEnabled(false);
-        model.callback.apply(generateEvent(), model.isEditing).whenComplete((o, throwable) ->  {
-            if(throwable != null){
-                Log.d(EventCreator.LOGCAT_TAG, "Event callback failed: " + throwable);
+    private boolean fieldsAreValid() {
+        String checkMsg = null;
+        if(model.name.getValue() == null || model.name.getValue().equals("")) {
+            checkMsg = String.valueOf(R.string.empty_event_name);
```
hugolan, line 238: why do you get the value of an empty event?

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
```diff
@@ -224,15 +232,45 @@ private Event generateEvent() {
...
+        } else if(model.name.getValue().length() > MAX_STRING_LENGTH) {
+            checkMsg = String.format(String.valueOf(R.string.event_name_too_long), MAX_STRING_LENGTH);
```
hugolan, line 240: is it the correct way to do this?
If an user puts a name that is too long and cutting it may not be the best solution why not ask the user to rewrite it?

hugolan (changes requested):

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
Flechman: What do you propose ? 50 ?

Flechman: We don't have much space issues so it's not a big problem to have long event names

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
Flechman: It's not the event that is empty, it's the name of the event. The variable name `empty_event_name` may not be the best, in fact it's a message to display to the user : `Please specify a name for the event.`
That way if the user didn't specify any name for his event, the system asks him to.

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
Flechman: Here we're not cutting it. If the event name is too long (that's the if condition), the user cannot create the new event and a message is displayed specifying that he should reduce the size of the event name.
So in fact here we are indeed asking the user to rewrite it.

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
hugolan: yes but maybe it is unrealistic, i doubt someone gonna name an event with 100 characters, maybe 50-70 should be good

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
hugolan: Ok I see

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
```diff
@@ -224,15 +232,45 @@ private Event generateEvent() {
...
+        } else if(model.startTime.getValue().toEpochSecond() > model.endTime.getValue().toEpochSecond()) {
+            checkMsg = String.valueOf(R.string.end_before_start);
+        } else if(model.endTime.getValue().toEpochSecond() - model.startTime.getValue().toEpochSecond() > Duration.ofDays(1).getSeconds()) {
+            checkMsg = String.valueOf(R.string.event_time_too_long);
+        } else if(model.customLocation.getValue() != null && model.customLocation.getValue().length() > MAX_STRING_LENGTH) {
+            checkMsg = String.format(String.valueOf(R.string.location_name_too_long), MAX_STRING_LENGTH);
+        } else if(model.isRecurrent.getValue()) {
```
TheImpressiveDonut, line 247: Shouldn't you check startTo,e, endTime, ... is not null like you did for the name?

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
hugolan: Sorry, I wasn't clear I meant to say that if you say to the user the name is too long he is going to cut his event name, we should better ask him to reformulate the name such that is it more concise instead of just saying you cannot enter names that are more than 100 characters.

TheImpressiveDonut (changes requested):

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
Flechman: I'll change to 50, 100 is indeed a lot for a title

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
Flechman: Hmmm if he  wants to cut  his  event name then so be it, it is him that chooses the name of his event. As long as his event  meets the constraints we should accept.

Flechman: + a "normal" user would naturally try to reformulate his event name if the name is too long.

hugolan: yes idk it was just a proposition, sometimes on apps you have messages like your password is not strong enough etc so I was just thinking maybe it could be could to suggest the user he should be concise with his names because the end goal is to be efficient to gain time

Flechman: What I could do is remove the current name that the  user put, but  maybe it would be annoying for him to have to retype everything. It's a matter of personal preference and I'm not sure A/B testing this feature now is a priority.

hugolan: yes sure but as the app is to gain time helping the user do it could be a great idea

app/src/main/java/com/github/onedirection/event/ui/MainFragment.java
Flechman: I wasn't sure if checking for null for the name was necessary, so I added it just in case. For the other attributes such as startTime, endTime, they have a default value when you're accessing the Event Creator window (ex: default value of  startTime is time.now()). Thus there's no need to check for null.

hugolan (approved):

TheImpressiveDonut (approved):

PR #228, opened 2021-05-16
Implement GCalendar events exporter

Ef55:


Ef55:
Note that, again, due to the high reliance of this code on systems out of our control, I won't be able to test this code.
(In particular, I couldn't find a way to log Cirrus' emulator into some google account, which would be strict minimum to even consider testing this code).

hugolan (approved):
good code not much to say here

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -39,9 +50,75 @@ public static Event toGCalendarEvents(com.github.onedirection.event.Event event)
         return null;
     }
 
+    private static List<String> listMatchingCalendarIds(Calendar service, String summary) {
+        List<String> ls = new ArrayList<>();
```
hugolan, line 54: maybe use a linkedlist here

Ef55: Why tho ?

hugolan: little bit faster in this case

Ef55: x_doubt

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -39,9 +50,75 @@ public static Event toGCalendarEvents(com.github.onedirection.event.Event event)
...
+                List<CalendarListEntry> items = calendarList.getItems();
+
+                for (CalendarListEntry calendarListEntry : items) {
+                    if (calendarListEntry.getSummary().equals(summary)) {
+                        ls.add(calendarListEntry.getId());
+                    }
+                }
+                pageToken = calendarList.getNextPageToken();
+            } while (pageToken != null);
+        } catch (Exception ignored) { /*Ignore and return empty*/ }
```
Sosol02, line 69: Please Log.d the exception so that we can still have a trace of it

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -39,9 +50,75 @@ public static Event toGCalendarEvents(com.github.onedirection.event.Event event)
...
+                        Log.d(LOGCAT_TAG, "Creating calendar...");
+                        String calendarId = service.calendars().insert(
+                                new com.google.api.services.calendar.model.Calendar()
+                                        .setDescription("Calendar generated by the 1Direction app.")
+                                        .setSummary(CALENDAR_SUMMARY)
+                        ).execute().getId();
+                        Log.d(LOGCAT_TAG, "... done (id " + calendarId + ")");
+
+
+                        assert calendarId != null;
```
Sosol02, line 104: Maybe turn that into Objects.requireNonNull, im not sure assertions are enabled in our builds.

Sosol02 (changes requested):

Sosol02:
Nice coverage too.

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
Ef55: Done.

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
Ef55: It was a leftover; should be gone now.

Sosol02 (approved):

PR #234, opened 2021-05-17
Event conversion into a Google Calendar Event

Flechman:


app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -18,19 +40,75 @@
     // Don't ask my the hows-and-whys of this string, all I know is that it doesn't work otherwise
     private static final String OAUTH_SCOPE = "oauth2:profile email";
     private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+    private static final String RFC3339_FORMAT = "yyyy-MM-dd'T'HH:mm:ssXXX";
```
hugolan, line 43: don't need the seconds right?

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -18,19 +40,75 @@
...
+     */
     public static Event toGCalendarEvents(com.github.onedirection.event.Event event) {
-        // TODO: Task #207 (assigned to @Flechman)
+        Objects.requireNonNull(event);
 
-        // Example: https://developers.google.com/calendar/v3/reference/events/insert#java
-        // Doc: https://developers.google.com/resources/api-libraries/documentation/calendar/v3/java/latest/
-        // Warning: use EventDateTime::setDatetime, not EventDateTime::setDate
+        Event gcEvent = new Event()
+                .setSummary(event.getName())
```
hugolan, line 69: why do you use setSummary instead of setName?

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -18,19 +40,75 @@
...
+                .setId(event.getId().getUuid());
 
-        return null;
+        if(event.isRecurrent()) {
+            Recurrence recurrence = event.getRecurrence().get();
+            if(!event.getId().equals(recurrence.getGroupId())) {
+                throw new IllegalArgumentException("Cannot convert a recurring event to a google event if this one is not the first element of the recurrence");
```
hugolan, line 75: why would you throw an exception in this case? To me if it the second occurrence of a recurring event it should be possible to transform it to a gc recurring as it is itself recurring. But maybe it is to hard to implement.

hugolan (changes requested):

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
Flechman: Because there's no method called setName hahaha, see the documentation just above that I've shared.
In Google Calendar "Summary" is the equivalent to our "Name"

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
Flechman: This is a "universal" format that  Google Calendar API uses, so I didn't choose this format. I didn't find any Java methods that use this format so I had to create this string.

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
Flechman: We should talk about it in vocal, I'm open to your suggestion. We've already talked about it with Noé and concluded this was the best way to do it.

hugolan (approved):

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -18,19 +40,75 @@
...
+                    recurrencePeriod = t.getValue();
+                }
+            }
+            if(recurrencePeriod == null) {
+                throw new IllegalArgumentException("The event recurrence period does not match any possible periods proposed.");
+            }
+
+            long count = (recurrence.getEndTime().toEpochSecond() - event.getStartTime().toEpochSecond()) / recurrence.getPeriod().getSeconds() + 1;
+
+            String[] recurr = new String[] {"RRULE:FREQ="+recurrencePeriod+";COUNT="+count};
```
Sosol02, line 91: I don't know if it is normal or not but u have written RRULE, maybe you wanted RULE instead? (I may be completely wrong here)

Sosol02 (approved):
Looks okay

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
Flechman: It's indeed RRULE, it's an existing format for the recurrence attributes of Google Calendar Events.

PR #235, opened 2021-05-18
Conversion from Google Calendar Event to Event

Flechman:
Please pay special attention to the logic to translate the recurrence, tell me what you would do better.

### DIFF
--- a/app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
+++ b/app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
@@ -73,7 +73,7 @@ public final class GoogleCalendar {
             for(String rule : recurrences) {
                 if(rule.substring(0, 5).equals("RRULE")) {
                     String[] info = rule.substring(11).split(";COUNT=");
-                
+
                 }
             }
         }
--- a/app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
+++ b/app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
@@ -3,6 +3,7 @@ package com.github.onedirection.interoperability.gcalendar;
 import android.accounts.Account;
 import android.content.Context;
 
+import com.github.onedirection.R;
 import com.github.onedirection.event.Event;
 import com.github.onedirection.event.Recurrence;
 import com.github.onedirection.utils.Id;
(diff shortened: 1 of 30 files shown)


app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -19,6 +25,7 @@
     private static final String OAUTH_SCOPE = "oauth2:profile email";
     private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
     static final String LOGCAT_TAG = "GCalendar";
+    private static final String DEFAULT_NAME = "No Name";
```
hugolan, line 28: why would the default name be no name? maybe put something generic there

hugolan (approved):
good code nothing much to say here

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
Flechman: What are your suggestions ? "No name" is good as it gives a hint to the user to give a real event name.

hugolan: I was thinking about something like insert name but ok

hugolan (approved):

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -111,10 +107,67 @@ public static Event toGCalendarEvents(com.github.onedirection.event.Event event)
...
+    public static Event fromGCalendarEvents(com.google.api.services.calendar.model.Event event) {
+        Objects.requireNonNull(event);
+        if(event.getStart() == null || event.getEnd() == null) {
+            throw new IllegalArgumentException("Event should have at least a start time and an end time.");
+        }
+
+        Id newId = Id.generateRandom();
+        String name = (event.getSummary() != null) ? event.getSummary() : DEFAULT_NAME;
+        //TODO: ask @Ef55 what default location name to put if no location is specified
+        String locationName = (event.getLocation() != null) ? event.getLocation() : "";
```
Ef55, line 128: I think that an empty string will do just fine.

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -111,10 +107,67 @@ public static Event toGCalendarEvents(com.github.onedirection.event.Event event)
...
+        if(event.getRecurrence() != null) {
+            List<String> recurrences = event.getRecurrence();
+            String periodically = null;
+            int eventCount = 0;
+
+            for(String rule : recurrences) {
+                if(rule.substring(0, 5).equals("RRULE")) {
+                    String[] info = rule.substring(11).split(";COUNT=");
+                    periodically = info[0];
+                    eventCount = Integer.parseInt(info[1]);
```
Ef55, line 146: - I think a little link to the doc wouldn't hurt (to answer questions like "why 11 ?" or "can info.length > 2 ?");
- I guess that no exception is thrown if multiple rules match, which might not be good.

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -111,10 +107,67 @@ public static Event toGCalendarEvents(com.github.onedirection.event.Event event)
...
+            if(periodically == null) {
+                throw new IllegalArgumentException("No recurrence rule matches the one used by "+ R.string.app_name+" Events");
+            }
+
+            Duration period = null;
+            for(Map.Entry<TemporalUnit, String> t : PERIODS.entrySet()) {
+                if(periodically.equals(t.getValue())) {
+                    period = t.getKey().getDuration();
+                }
+            }
```
Ef55, line 158: I think that this map could be put to better use:

- To do _TemporalUnit_ -> _String_, I guess that _.toString()_ might be sufficient;
- The key/values could then be reversed, and you could then simply do `PERIODS.get(periodically).getDuration()`.

Ef55 (changes requested):
Just a few details to fix, the overall logic seems good.

I would also strongly encourage you to add a "coherence test", where you convert an event into its GCalendar form, and then convert it back into our Event, to final test equality with the original event.

Flechman:
> Just a few details to fix, the overall logic seems good.
> 
> I would also strongly encourage you to add a "coherence test", where you convert an event into its GCalendar form, and then convert it back into our Event, to final test equality with the original event.

My test on the conversion from GCalendar to Event does this check already. I cannot test for equality on events since the conversion changes the ids (because GCalendar id formatting may not be the same as UUID).

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
Flechman: Done

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
Flechman: I don't understand your first point, what is it for ? 
Note that the String is a period such as "DAILY", not a  technical duration format

Flechman: Second point is done

Ef55: You're indeed right; didn't realized that the strings weren't simply the name of the period.

Ef55:
> My test on the conversion from GCalendar to Event does this check already. I cannot test for equality on events since the conversion changes the ids (because GCalendar id formatting may not be the same as UUID).

This might be slightly annoying when we will import events back from GCalendar. 
We will have to discuss that with the «Product owner», and do changes (in a subsequent PR) accordingly.

Ef55 (approved):

PR #263, opened 2021-05-21
Fix string references for the error messages when sanitizing user inputs in event creator

Flechman:


Flechman:
I merged with the work of @Ef55, and since he couldn't test his work this explains why I only have 2% diff-coverage. See PR #228 for explanations as to why it couldn't be tested.

justagoodguy-omega (approved):

Ef55 (approved):

PR #265, opened 2021-05-22
Cleanup package structure

Flechman:


Flechman:
Screen Shot 2021-05-22 at 12 10 43 PM

Here is the final package tree structure. I invite you to look at the package structure yourself on this branch, and look at the files in each package.

Ef55:
Overall it's really nice, but here I go with some quality bikesheding:

- I'm not a huge fan of having a package named _database_ inside the _database_ package. I would suggest  _implementation_ as a possible change.
- I think that in some cases, it's a bit too much, in particular when a package ends up with only one file, such as _navigation.activity_ or _database.queries_. 

Also, since this PR is about package naming, would it be ok for you if I renamed _event.ui_ to something a bit more precise, like _event.creator_ ?

Flechman:
> Overall it's really nice, but here I go with some quality bikesheding:
> 
> 
> 
> - I'm not a huge fan of having a package named _database_ inside the _database_ package. I would suggest  _implementation_ as a possible change.
> 
> - I think that in some cases, it's a bit too much, in particular when a package ends up with only one file, such as _navigation.activity_ or _database.queries_. 
> 
> 
> 
> Also, since this PR is about package naming, would it be ok for you if I renamed _event.ui_ to something a bit more precise, like _event.creator_ ?

For first point, i'll make the change. For second point, i think it's okay to have packages with a single class in it, but if it really really bothers you, i let you make the changes. For last point it's ok.

Ef55:
Thanks; I did my change.
I just removed the _navigation.activity_ package, since activity won't ever have another activity.
I'll accept the PR as soon as you did you change.

Sosol02 (commented):
After the comments made by @Ef55 are accepted i will approve as well.

Sosol02:
Also there is a test that's failing because of the move :)

Sosol02 (approved):

Ef55 (approved):

Issue #268, opened 2021-05-22
GCalendars fails on recurring events.


Ef55:
Conversion (rightfully) fails on non-root recurring events. 
Such events should be filtered out.

Issue #266, opened 2021-05-22
Generated GCalendar events' IDs are invalid


Ef55:
The generated IDs do not follow the [guidelines](https://developers.google.com/calendar/v3/reference/events/insert). Here is the relevant passage:

```
Opaque identifier of the event. When creating new single or recurring events, you can specify their IDs. Provided IDs must follow these rules:

   - characters allowed in the ID are those used in base32hex encoding, i.e. lowercase letters a-v and digits 0-9, see section 3.1.2 in RFC2938
   - the length of the ID must be between 5 and 1024 characters
   - the ID must be unique per calendar

Due to the globally distributed nature of the system, we cannot guarantee that ID collisions will be detected at event creation time. To minimize the risk of collisions we recommend using an established UUID algorithm such as one described in RFC4122.

If you do not specify an ID, it will be automatically generated by the server.

Note that the icalUID and the id are not identical and only one of them should be supplied at event creation time. One difference in their semantics is that in recurring events, all occurrences of one event have different ids while they all share the same icalUIDs.
```
This prevent events export.

PR #270, opened 2021-05-22
Fix IDing and recurrence handling in GCalendar export

Ef55:
From #268 
```
Conversion (rightfully) fails on non-root recurring events.
Such events should be filtered out.
```

From #266 (go on dedicated issue for more details)
```
The generated IDs do not follow the guidelines [from GCalendar]. [...]
```


Ef55:
Note: awful coverage doesn't come from those changes, but from the previous coverage of the changed class.

Sosol02 (approved):
Looks good, not much to say since this is a fix.

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -53,18 +60,42 @@
...
     private GoogleCalendar() {
     }
 
+    /** Remove '-' from Id, as they are not allowed by GCalendar: https://developers.google.com/calendar/v3/reference/events/insert */
+    private static String fromId(Id id){
+        return id.getUuid().replace("-", "");
+    }
+
+    private static Id toId(String str) {
+        if (str.length() != Id.LENGTH - Id.SEP_COUNT) {
```
Flechman, line 77: I assume this method will be called for converting GCalendar ids into Ids. Is it certain that the id format of all GCalendar ids verify this condition ?

Ef55: Well, we are the one setting the IDs when exporting, so everything should be good.
Of course, if the user starts adding events into the calendar, then we don't control the IDs anymore, but that should be handled somewhere else.

Flechman (approved):
Thanks for rewritting the tests

PR #272, opened 2021-05-23
General small fixes + code review by coaches

Flechman:
+ Fix incorrect calendar dates
+ Apply the strict minimum documentation
+ Remove the most important Lint warnings (there are still some, either because they are false positive, or because they are not really relevant)
+ In event.ui.ViewModel, I couldn't manage to put attributes in private and make getters, because the app crashes.


app/src/main/java/com/github/onedirection/database/implementation/CachedDatabase.java
```diff
@@ -163,10 +163,9 @@ public int hashCode() {
         }
     }
 
-    // in these functions most null checks are done in the Query constructor.
-
     @Override
     public <T extends Storable<T>> CompletableFuture<List<T>> filterWhereEquals(String key, Object value, Storer<T> storer) {
+        // 'null' checks are done in the Query constructor.
```
Ef55, line 168: This comment is (imo) useless.

app/src/main/java/com/github/onedirection/database/utils/FirebaseUtils.java
```diff
@@ -4,6 +4,9 @@
 import com.google.firebase.firestore.FirebaseFirestore;
 import com.google.firebase.firestore.FirebaseFirestoreSettings;
 
+/**
+ * Utility class for using Firebase 'Firestore' database tools
+ */
```
Ef55, line 9: I opened issue #278 regarding this lint warning.

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -47,7 +46,7 @@
  */
 public final class GoogleCalendar {
 
-    // Don't ask my the hows-and-whys of this string, all I know is that it doesn't work otherwise
+    // Required string ; should not be changed
```
Ef55, line 49: This string could (and most likely **should**) be changed, but the person would have to understand all the interactions between _oAuth2_, _GoogleSignIn and_ _Google Calendar._

Ef55 (approved):
First of all, thanks for fixing all those little things.
I just think you should have done this in multiple PRs:

- This is a pain to review;
- This is gonna be hell for anyone who created a branch before this one, but will merge after this one is merged into main.

app/src/main/java/com/github/onedirection/navigation/fragment/map/MapFragment.java
```diff
@@ -72,17 +72,16 @@
     private ActivityResultLauncher<String> requestPermissionLauncher;
     private CompletableFuture<Boolean> permissionRequestResult;
 
-    private Button navigationButton;
```
Sosol02, line 75: Why did you remove the button and put it in the method? The goal of putting the buttons and views in the class is to avoid having to call findViewById multiple times, and even though the navigation is not in use in other places now, it is more consistent to keep it the way it was i think.

Sosol02 (changes requested):
Looks good overall!

app/src/main/java/com/github/onedirection/navigation/fragment/map/MapFragment.java
Flechman: Ok then I'll undo this change. I made it in the first place because it was recommended by the static code analysis.

Flechman:
> First of all, thanks for fixing all those little things.
> I just think you should have done this in multiple PRs:
> 
> * This is a pain to review;
> * This is gonna be hell for anyone who created a branch before this one, but will merge after this one is merged into main.

Yes I should've made multiple smaller PRs. The changes are very small but they are made on multiple files each time.

To prevent from being hell for anyone who will have to merge, I should've done these fixes after the work of everyone (at the end of the week). I had to do most of my work of this week last week-end because I have exams this week. I hope the team understands and can deal with the potential unpleasant merges.

Sosol02 (approved):

Issue #271, opened 2021-05-22
From GCalendar should return multiple events (in case of recurrent GCalendar Event).


Ef55:

Ef55:
Fixed using another method. See #284

PR #277, opened 2021-05-26
Setup GCalendar import

Ef55:
Another GCalendar PR, another terrible day for the test coverage.

### DIFF
--- a/app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
+++ b/app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
@@ -407,7 +407,7 @@ public final class GoogleCalendar {
                 Log.d(LOGCAT_TAG, newEvents.size() + " g-events to add to DB.");
                 EventQueries adder = new EventQueries(Database.getDefaultInstance());
 
-                for (Event event : gcalendar) {
+                for (Event event : newEvents) {
                     if (event.isRecurrent()) {
                         adder.addRecurringEvent(event);
                     } else {
--- a/app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
+++ b/app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
@@ -6,13 +6,15 @@ import android.util.Log;
 
 import androidx.annotation.VisibleForTesting;
 
-import com.github.onedirection.BuildConfig;
 import com.github.onedirection.R;
+import com.github.onedirection.database.implementation.Database;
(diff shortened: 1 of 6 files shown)


Ef55:
**Note:** currently, import of recurring events doesn't work properly (the event is not recurrent).
This is mostly related to either #269 or #271. I will look at this next thing (so not in this PR).

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -312,5 +351,74 @@ public static Event fromGCalendarEvents(com.google.api.services.calendar.model.E
...
+                for (Event event : gcalendar) {
+                    if (!database.contains(event.getId())) {
+                        newEvents.add(event);
+                    }
+                }
+
+                Log.d(LOGCAT_TAG, newEvents.size() + " g-events to add to DB.");
+                EventQueries adder = new EventQueries(Database.getDefaultInstance());
+
+                for (Event event : gcalendar) {
```
Flechman, line 410: Instead of iterating over `gcalendar`, aren't you supposed to iterate over `newEvents` ?

Flechman (changes requested):
Well written import logic, you managed to make only 1 query.
I think you made a small mistake in the import logic, so I will wait for you to  view my comment on it before approving.

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
```diff
@@ -100,51 +107,56 @@ private static Id toId(String str) {
...
+            gcEvent.setStart(start);
+
+            DateTime endDateTime = new DateTime(event.getEndTime().format(DateTimeFormatter.ofPattern(RFC3339_FORMAT)));
+            EventDateTime end = new EventDateTime()
+                    .setDateTime(endDateTime)
+                    .setTimeZone(ZoneId.SHORT_IDS.get(event.getEndTime().getZone().getId()));
+            gcEvent.setEnd(end);
+
+            return gcEvent;
+        } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
Ef55: Before, exceptions would fail silently at call point (since it is wrapped inside a future).
I added this as a debug help.

app/src/main/java/com/github/onedirection/interoperability/gcalendar/GoogleCalendar.java
Ef55: Should be fixed.

Flechman (approved):

justagoodguy-omega (approved):
Nice code looks good!

PR #279, opened 2021-05-26
Add support for Geolocated events in GCalendar interop and fix recurring events import/export

Ef55:


Flechman (approved):
Code seems okay. It's too bad that we didn't manage to store in a more proper way the recurring events. But the api isn't really clear enough as to how to use it properly.

justagoodguy-omega (approved):
Yep can merge

PR #295, opened 2021-06-02
Add package-info to most classes

Ef55:
Note that not all packages are yet documented.


app/src/main/java/com/github/onedirection/startup/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * Code called at application start up to initialize some functionality.
```
hugolan, line 2: could be more precise here

hugolan (approved):
good description of the classes

app/src/main/java/com/github/onedirection/startup/package-info.java
Ef55: I personally don't have much more to say about this package.
If you have any suggesting, I would be more than happy to include them.

TheImpressiveDonut (approved):
Unbelievable documentation. It was so good that I fell from my chair !

PR #299, opened 2021-06-03
Change geolocation checkbox behavior to be less surprising

Ef55:
Change the behavior of the "geolocation checkbox".
Instead of opening the geolocation tab, the event is not created until it is set.


hugolan (approved):
good job finally someone didn't forgot to do it haha :)

justagoodguy-omega (approved):

PR #301, opened 2021-06-03
Fix fast navigation between map and others fragment

TheImpressiveDonut:


hugolan (approved):
good  code but maybe don't forget to test it also even though it is not too long

Ef55 (approved):

PR #302, opened 2021-06-03
fix display drawer for fragment

TheImpressiveDonut:


Ef55 (approved):
Wow, many code, much improvement.

hugolan (approved):
Wow, many code, much improvement.

